PR #10, opened 2015-10-21
Update PusherSwift.swift

daande:
Calls can throw but not caught errors in iOS 9.0.


hamchapman:
Thanks! Seems like a good idea in principle. Although if the notifier can't be started I'm not sure it makes sense to print that the network in unreachable. 

Also, you can see that the tests are now failing on Travis. Any chance you could take a look at making sure that they pass?

daande:
@hamchapman Just tried changing some things lets see if it builds!

hamchapman:
Fixed with latest version (0.1.3)

PR #12, opened 2015-10-30
reachability 2.1 support + options for host and port

tchukuchuk:
Hi,

I have added the possibility to set host and port as options to use different server than pusher. I need it to use slanger.

I have also added the support for ReachabilitySwift 2.1.

Best regards,

Julien


hamchapman:
This looks awesome in principle! Will give it a look over the next few days, when I get a chance, and get it merged as soon as possible :+1:

hamchapman:
If `encrypted` is true the port needs to be set as 443 so taking it from the options isn't really required / helpful I don't think.

That's causing the breaking test at the moment, otherwise this is looking like a good set of changes.

hamchapman:
This is great - sorry it took me a while to merge! Thank you!!

PR #19, opened 2016-01-02
Update podspec to support tvOS

goose2460:


hamchapman:
This is great! Last time I checked Reachability didn't support tvOS but it looks like it does now :+1: 

Only tiny thing is that it looks like you've set the dependency to be CryptoSwift 2.1 but it's actually using https://github.com/krzyzanowskim/CryptoSwift/releases/tag/0.2.2)

If you can check that then I'll get it merged in and publish 0.1.7 :)

goose2460:
Whoops, should be good now!

hamchapman:
Yeah that's all fine :+1:

PR #31, opened 2016-03-10
Send the "pusher:subscription_succeeded" event

bdolman:
This makes it possible to implement a [message history use case](https://blog.pusher.com/how-to-add-message-history-to-your-pusher-apps/) like that outlined on the Pusher website. 

I used the pusher [JS library](https://github.com/pusher/pusher-js) for reference while implementing.


hamchapman:
Again, looks good to me :+1: 

I'll cut a new release later today

PR #32, opened 2016-03-10
Fix autoreconnect

bdolman:
The reachability object was was not being retained, so it was being
deallocated right after being created.

This also stops the autoreconnect behavior if the connection is explicitly
disconnected.


hamchapman:
This is awesome - thanks @bdolman !

PR #36, opened 2016-04-04
Add CryptoSwift directly, just HMAC sha256 parts

hamchapman:
As the title says


zimbatm:
Make sure the license is updated appropriately

PR #42, opened 2016-05-05
Update README.md - added authenticated channel example

xeo-it:
Added an example for subscribing to authenticated channel.  
I was a little bit confused since in the official pusher website, the example is in objective-c and it uses a delegate so it's a bit different.  
Hopefully this will be helpful and avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,27 @@ let pusher = Pusher(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


hamchapman:
Awesome - thanks! 

I was thinking about adding something like this in but you've done it for me 👍

PR #47, opened 2016-05-27
Pass Authorization Errors Down to Client

psycotica0:
Currently we're just logging the error, but I need that at the client side.
I thought about a delegate, but I figured in this case we have a channel that's
listening to events, so I'll just send an event out.

Thoughts?

### DIFF
--- a/Source/PusherConnection.swift
+++ b/Source/PusherConnection.swift
@@ -302,6 +302,24 @@ public class PusherConnection {
     }
 
     /**
+        Handle failure of our auth endpoint
+
+        - parameter data: The error returned by the auth endpoint
+    */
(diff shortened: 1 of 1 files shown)


psycotica0:
I made this PR against `split-up-and-document` because that's what we're using in our app, and I didn't want to move upstream too much.

Feel free to merge it around if preferred.

hamchapman:
Nope, that's cool. Hopefully going to get a release based on the `split-up-and-document` branch out very soon.

And thanks a lot for this! Looks good to me 👍

hamchapman:
I will probably add a test for it before releasing but I'm good to merge it for now.

PR #34, opened 2016-04-03
[WIP] 1.0.0 release

hamchapman:
- [x] Split up PusherSwift into multiple files
- [x] Split up tests into multiple files
- [x] Bring CryptoSwift, Starscream and Reachability dependencies inside the library
- [x] Add documentation to code
- [x] Add `onMemberAdded` and `onMemberRemoved` properties to presence channel
- [x] Add `findMember` and `me` functions to presence channel
- [x] Change connection change delegate based on comments in #50 
- [x] Add ability to have logging for all messages, as in #45 
- [x] Investigate handling client errors better #52
- [x] Clean up cocoapods-related code
- [x] Check carthage still working
- [x] Final checks

PR #49, opened 2016-05-31
Dispatch Errors on Main Thread

psycotica0:
The other handlers are called by Starscream's delegate, but the auth
errors are called by `NSUrlSession`, which is on a background thread.

So, I've moved the logic into a `dispatch_async` on the main thread.
I decided to put it in the handleError method rather than outside it because
there are multiple places that need it and currently all of them are on the
background threads.

If later that's not true, then we should put the dispatch around the method
call in all 4 or so places it comes up.

### DIFF
--- a/Source/PusherConnection.swift
+++ b/Source/PusherConnection.swift
@@ -314,7 +314,9 @@ public class PusherConnection {
             "channel": channelName,
             "data": data ?? ""
         ]
-        handleEvent(eventName, jsonObject: json)
+        dispatch_async(dispatch_get_main_queue()) {
+            self.handleEvent(eventName, jsonObject: json)
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


hamchapman:
This looks sensible, although I just committed some code that makes some changes to the bit you changed in this PR: https://github.com/pusher/pusher-websocket-swift/pull/34/commits/77bd1eaf8c588cb72002ea4b92f7ec4fbe499a0c

If you could rebase that would be great 👍

hamchapman:
@psycotica0 have you had a chance to look at this with the changes?

psycotica0:
No, sorry. Doesn't seem like a problem, I've just been tied up, sorry.

hamchapman:
Awesome - thanks a lot! And no worries  👍

PR #54, opened 2016-06-16
Feature/authrequestcustomizer improvements

Noobish1:
Passing a `NSMutableURLRequest` into the `authRequestCustomizer` isn't very helpful when the customizer needs the socket, so doing it this way gives the customizer everything it needs.

### DIFF
--- a/Source/PusherClientOptions.swift
+++ b/Source/PusherClientOptions.swift
@@ -16,7 +16,7 @@ public struct PusherClientOptions {
     public let host: String?
     public let port: Int?
     public let autoReconnect: Bool?
-    public let authRequestCustomizer: (NSMutableURLRequest -> NSMutableURLRequest)?
+    public let authRequestCustomizer: ((endpoint: String, socket: String, channel: PusherChannel) -> NSMutableURLRequest)?
     public let debugLogger: ((String) -> ())?
 
     /**
@@ -89,7 +89,7 @@ public struct PusherClientOptions {
         self.host = optionsMergedWithDefaults["host"] as? String
         self.port = optionsMergedWithDefaults["port"] as? Int
         self.autoReconnect = optionsMergedWithDefaults["autoReconnect"] as? Bool
-        self.authRequestCustomizer = optionsMergedWithDefaults["authRequestCustomizer"] as? (NSMutableURLRequest -> NSMutableURLRequest)
+        self.authRequestCustomizer = optionsMergedWithDefaults["authRequestCustomizer"] as? ((String, String, PusherChannel) -> NSMutableURLRequest)
         self.debugLogger = optionsMergedWithDefaults["debugLogger"] as? (String) -> ()
 
         if let _ = authEndpoint {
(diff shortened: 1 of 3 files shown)


hamchapman:
I'm not against this but what's the use case here? Currently the socket is already set as part of the HTTP body on the request, before the request is passed to the customiser. 

Does this not work for your use case?

Noobish1:
The endpoint we use for authentication requires different parameter names so the current method doesn't work well for us. If we don't have this change we'd have to parse the socket out of the body which is gross.

hamchapman:
Parsing the socket out of the body shouldn't be a problem at the minute though. For example, you can do it like this using a little Ruby (SInatra) app:

```ruby
post '/pusher/auth' do
  content_type :json

  if params[:channel_name].include?('private')
    Pusher[params[:channel_name]].authenticate(params[:socket_id]).to_json
  else
    Pusher[params[:channel_name]].authenticate(params[:socket_id], {
      user_id: user_id,
      user_info: {
        time: Time.now
      }
    }).to_json
  end
end
```

Is there a reason that you can't access the `socket_id` on your server when the auth request arrives?

Again, I'm not necessarily against making the change, but it is a breaking change, and would require more setup for users of the library in that if they want to, for example, just add a token to the request, then in creating an `authRequestCustomizer` they'd then also have to make sure the socket and channel name values are set.

Noobish1:
We can't change our backend as we have a website using the same endpount for pusher authentication atm. I realise this is a breaking change. When I said parsing the socket from the HTTPBody, I meant doing that on the swift side (as we can't change the backend), which means we'd have to turn the NSData into a string, split it based on the ampersands then split it on the equals signs. As for clients doing more work, you benefit from being open source in that they can just see how the library makes the request. Another option is to pass a pre-made request as well.

hamchapman:
Thanks for the clarifications. All sounds reasonable to me.

I've created a new PR: https://github.com/pusher/pusher-websocket-swift/pull/56

It marks the `authRequestCustomizer` as deprecated, but leaves it in for now, and adds an `authRequestBuilder`, which is very similar to your PR, except instead of a channel it takes just the channel name.

Noobish1:
`authRequestBuilder` works for me. I do have another comment though: would you be against moving the `authRequestBuilder` to the `PusherConnection`?, having it in the options leads to pain initializing `Pusher`.

Noobish1:
Same goes for the rest of the closures on the `PusherClientOptions`

hamchapman:
Again, I'm not necessarily against it, but what sort of pain do you mean?

Noobish1:
if you want to reference `self` in any of the closures you have to do more lazy initialization to get around it.

I've made a bunch of changes to the `PusherClientOptions` in a fork, you can see the diff [here](https://github.com/pusher/pusher-websocket-swift/compare/pusher:master...Noobish1:feature/pass-all-params-to-authrequestcustomizer) if you want to have a look. Using a dictionary for options makes you do a whole heap of validation that you shouldn't have to.

hamchapman:
Good point.

And that fork looks like it contains some nice changes! Want to make it into a PR? Looks like it might involve some tests being changed and might need some tweaking. But definitely has some good parts to it

Noobish1:
Yea I've definitely broken a heap of your tests, i'm sorry about that :p (party why I just PRed this change by itself).

Sure I can PR the whole thing.

Noobish1:
whole thing is in #57, I renamed authRequestCustomizer -> authRequestBuilder quickly as well.

hamchapman:
Replaced by #60

Issue #43, opened 2016-05-06
disconnecting while entering the background


mrlantan:
I recently found, that when I hit the power button, the device unsubscribes from the channel and when the app comes back to foreground it doesn't connect again.
I decided to do it manually in connectionChange delegate method in the way
```objective-c
func connectionChange(old: ConnectionState, new: ConnectionState) {
        switch new {
        case .Disconnected:
                self.pusher.connect()
        }
        ...
}
```

This didn't work because in your
 ```objective-c
private func updateConnectionState(newState: ConnectionState) {
        self.stateChangeDelegate?.connectionChange(self.connectionState, new: newState)
        self.connectionState = newState
}
```
you firstly call the delegate method and then change the state. So when i tried to call
```objective-c
self.pusher.connect()
```

```objective-c
public func connect() {
        if self.connectionState == .Connected {
            return
        } else {
            updateConnectionState(.Connecting)
            self.socket.connect()
            if let reconnect = self.options.autoReconnect where reconnect {
                _ = try? reachability?.startNotifier()
            }
        }
    }
```

this func returned because current state was .Connected

I would suggest swapping lines in updateConnectionState method  creating a variable to store the self.connectionState value and pass it then to the delegate method parameters. So it may look like :
```objective-c
private func updateConnectionState(newState: ConnectionState) {
    let oldConnectionState = self.connectionState
    self.connectionState = newState   
    self.stateChangeDelegate?.connectionChange(oldConnectionState, new: newState)
}
```

hamchapman:
Interesting - was it reconnecting successfully but then not resubscribing to any channels you had been subscribed to?

And that's a very valid point in terms of swapping the order of the lines - do you want to make a PR for it?

mrlantan:
So here is my flow:
1. I run the app and pusher connects and subscribes to a channel.
On the dashboard.pusher.com I get these lines
![image](https://cloud.githubusercontent.com/assets/8770111/15070051/923110ba-138a-11e6-97cf-78fe0fcbdb43.png)
2. I hit the power button on my IPhone and immediately get this line
![image](https://cloud.githubusercontent.com/assets/8770111/15070068/adbe3858-138a-11e6-99b3-83b88946993f.png)
3. I unlock the iphone, bring the app back to foreground and when I try to send pushes, callbacks are not called.

I worked around an issue that I described in the post by calling connect in this way
```objc
case .Disconnected:
            dispatch_async(dispatch_get_main_queue(), {
[...]

PR #55, opened 2016-06-16
More reliable reconnect

psycotica0:
I'm interested to hear your thoughts on this.

Currently there's no reconnect logic except to connect again when the network becomes reachable.
That's fine if the websocket only closes due to network unreachability, but there are other cases that also might happen.

The remote end could close the connection probably, but the one I saw was that the app goes to sleep in the background, and 20 minutes later comes back to the foreground.

In this case everything is still like it was, but the websocket is closed.
Reachability doesn't fire, though, because reachability hasn't changed.

So, this sets up a reconnect when the websocket dies, but cancels it if we hear from reachability for any reason.
Thoughts?

### DIFF
--- a/Source/PusherConnection.swift
+++ b/Source/PusherConnection.swift
@@ -17,15 +17,18 @@ public class PusherConnection {
     public var socket: WebSocket!
     public var URLSession: NSURLSession
     public weak var stateChangeDelegate: ConnectionStateChangeDelegate?
+    internal var reconnectOperation: NSOperation?
 
     public lazy var reachability: Reachability? = {
         let reachability = try? Reachability.reachabilityForInternetConnection()
         reachability?.whenReachable = { [unowned self] reachability in
+            self.reconnectOperation?.cancel()
             if self.connectionState == .Disconnected {
                 self.socket.connect()
             }
         }
         reachability?.whenUnreachable = { [unowned self] reachability in
+            self.reconnectOperation?.cancel()
             print("Network unreachable")
         }
         return reachability
--- a/Source/PusherWebsocketDelegate.swift
+++ b/Source/PusherWebsocketDelegate.swift
@@ -40,16 +40,18 @@ extension PusherConnection: WebSocketDelegate {
             channel.subscribed = false
         }
 
-        let operation = NSBlockOperation {
-            self.socket.connect()
-        }
+        if let reconnect = self.options.autoReconnect where reconnect {
+            let operation = NSBlockOperation {
+                self.socket.connect()
+            }
 
-        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, Int64(NSEC_PER_SEC)), dispatch_get_main_queue()) {
-            NSOperationQueue.mainQueue().addOperation(operation)
-        }
+            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, Int64(NSEC_PER_SEC)), dispatch_get_main_queue()) {
+                NSOperationQueue.mainQueue().addOperation(operation)
+            }
 
-        self.reconnectOperation?.cancel()
-        self.reconnectOperation = operation
+            self.reconnectOperation?.cancel()
+            self.reconnectOperation = operation
+        }
     }
 
     public func websocketDidConnect(ws: WebSocket) {}
--- a/Source/PusherWebsocketDelegate.swift
+++ b/Source/PusherWebsocketDelegate.swift
@@ -41,16 +41,18 @@ extension PusherConnection: WebSocketDelegate {
         }
 
         if let reconnect = self.options.autoReconnect where reconnect {
-            let operation = NSBlockOperation {
-                self.socket.connect()
-            }
+            if let reachability = self.reachability where reachability.isReachable() {
+                let operation = NSBlockOperation {
+                    self.socket.connect()
+                }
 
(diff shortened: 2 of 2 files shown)


hamchapman:
Just to clarify the scenario you encountered:

- Websocket is connected with an app that is foregrounded
- App gets backgrounded
- Eventually websocket gets closed but reachability not triggered
- App is foregrounded again but websocket connection is closed and reachability has no reason to cause a reconnect

Is that correct?

If that's correct, is this PR designed as a fix to make sure that a reconnect occurs? 

I can't quite figure out what's going on with the code. For example, why when reachability receives a notification that the network is reachable does the `reconnectOperation` get sent a `cancel()`?

I'm not massively familiar with `NSOperation` so I may just not know the API well enough to know what's going on.

Thanks! I'm definitely keen to get this merged if it fixes things, just once I've got my head around it 👍

psycotica0:
Yeah, that's the scenario.
I'm specifically on tvOS, and the steps I took to reproduce the bug was to run my app, then leave my app and wait 15 minutes for the tvOS to go to sleep (when it was set to go to sleep after15 minutes).

Then, wake the system up and go back into my app.

But, more generally, I believe it could come up if the underlying socket closed for any reason.
Like, if pusher hiccuped, or maybe redeployed, and closed the socket on their side.
Or, I think, if the wifi you're connected to dropped packets for a bit.
Reachability wouldn't trigger, because you're still connected, but the websocket would still timeout, I think.

I didn't test that one though.

So, in terms of implementation I'm not super proud of this one myself.
Here's what I was going for:

If the websocket goes down for reachability reasons, then the code is already solid as-is.
But, in my testing it seemed that the websocket closed before reachability had changed.

So, I used NSOperation because it's cancelable, which means that if I've called `cancel()` on it, then the block I've constructed it with will not be executed once it's enqueued.

So, when the websocket disconnects it makes a block to reconnect, and waits 1 second to do it.
If, in that 1 second, reachability changes for any reason, then I cancel the operation to allow the logic that was already there and working to do its thing.

If reachability goes down, then I don't want to connect, because it won't work.
If reachability goes up, then I don't want my delayed connect, because I'm going to call connect right now.

But, if after 1 second nothing has canceled the operation, then that means that the websocket disconnected for a reason unrelated to reachability changes, so we want to reconnect.

As a fun side effect, if Pusher's server runs into trouble and dies temporarily, it also means that there's a 1 second delay between retries, which is often nice. No reason to spam them with as many retry attempts as the CPU will allow.

There are other ways to write the kind of logic I've just described, but I chose NSOperation because of `cancel()` keeps its own state, which makes things cleaner in the case where reachabiity goes up and down more than once in that second. I won't accidentally spin up 10 overlapping `delay_after` blocks that all do who knows what by the time the first runs.

But, this is your codebase, and if you'd prefer it to be structured another way, then I'd be willing to hear it.

psycotica0:
Also, just looked through the closed issues. Do you think this might have something to do wth #43?

hamchapman:
Ah I see! Yeah, that makes a lot of sense. 

Have you tested it with your tvOS app to see if it works in a real-world setting?

For example, if the Apple TV goes to sleep will the code where the reconnect operation is added to the operation queue actually get run? Or if the device goes to sleep will the websocket close and the code in the websocket delegate just not get run?

I can see how this definitely helps the case where the connection is closed by the server as opposed to the client though, so it's an improvement nonetheless 👍

psycotica0:
Yes, the testing loop is kinda long because I have to set it up then wait 20 minutes, but I spent most of yesterday testing it, and I don't know that it catches every case but it at least catches all of the ones I've seen so far. 

Even before I added my fix, I always saw the connection delegate properly update the status to disconnected, so that gives me faith. 

I'm going to keep playing with it today to see if I can get it to break any other ways.

hamchapman:
Awesome - thanks a lot for doing all that testing and reporting back. The pull request is a huge bonus as well!

Let me know if you hit anything else today. I'll merge this very soon and get a new release out.

psycotica0:
Not a problem. It's not like I don't also benefit from the library being more resilient on my behalf!

Just woke the app up from an overnight sleep, and it connected basically instantly.
Still more testing to do, but I continue to be happy.

psycotica0:
Ok, small bug. I've tested a bunch of cases and most seem like what I would expect, but if I start the app with no reachability then it attempts to reconnect once per second (which fills my log)

So, I might have to do a small tweak to handle the "started with no network case".

The app still works, and connects when network shows up, and functions fine after that in both the disconnect and reconnect phases, but it's not ideal at the start.
Planning my approach now.

psycotica0:
Ok, so, I think I've fixed that issue as well. The first time (or theoretically any time) that I know the network is unreachable when the socket disconnects, don't even both going through the new retry logic.

The assumption being that if the reachability stuff thinks the network is unreachable, then when it becomes reachable, it will tell me.

This shouldn't effect the code-path I'd previously written (and in my tests it appears to work in the same cases), but on initial connection in a state of disconnection it doesn't retry uselessly over and over, which was a minor regression I'd caused.

I did some more testing today and haven't found any regressions, and it appears to be more reliable in the cases I've previously identified.

hamchapman:
Looking great! I'm going to bundle this up with the changes in #60 and do a 2.0.0 release.

PR #58, opened 2016-06-20
[WIP] Native push

jameshfisher:


Source/NativePusher.swift
```diff
@@ -0,0 +1,157 @@
...
+    private var clientId: String?
+    
+    // Queued actions to perform when the client is registered.
+    private var outbox: Array<(String, SubscriptionChange)>
+    
+    // Expected to be initialized by instances of Pusher.
+    // Normal clients should access the shared instance via Pusher.nativePusher().
+    public init(pusherAppKey: String) {
+        self.pusherAppKey = pusherAppKey
+        clientId = nil
```
hamchapman: You can set these as default values inline, like this: `private var clientId: String? = nil`

PR #57, opened 2016-06-17
Feature/pusherclientoptions improvements

Noobish1:
I've done a number of things in order to cleanup the `PusherClientOptions` and make it more Swifty™.

PR #60, opened 2016-06-21
Noobish1/improvements

hamchapman:
Based on #57 

I just fixed up some stuff aesthetically and fixed (most) of the tests. 2 still failing but easily fixable. I just stopped because I'm not sure that the `PusherClientOptions`, as it is, is the nicest way for a developer to interact with the library in terms of setting options. It's certainly far Swiftier but if, for example, you just want to set `encrypted` to false, then you have to do something like this:

```swift
let options = PusherClientOptions(
    authMethod: .NoMethod, 
    attemptToReturnJSONObject: true, 
    autoReconnect: true, 
    host: PusherHost.Cluster("mt1"), 
    port: nil, 
    encrypted: false
)
```

which seems sub optimal.

@Noobish1 WDYT?

-- Update --

You don't have to do that ^^ 👍


Noobish1:
are you saying the options are bad because you have to pass in multiple options? because you don't have to do that because most of those things have default values.

hamchapman:
Replying on my phone so I can't check but I thought you still had to specify the values for parameters that come before a later one, i.e. can you do:

```swift
let options = PusherClientOptions(encrypted: false)
```

Noobish1:
yup you only have to specify what you need or the stuff that doesn't have default values (which is nothing in this case). You can do `let options = PusherClientOptions()` even.

Noobish1:
That's actual the default value for the options in the `Pusher` init method.

Noobish1:
@hamchapman: Got another possible tweak to the auth builder: https://github.com/Noobish1/pusher-websocket-swift/compare/1ca0df3...Noobish1:feature/pass-all-params-to-authrequestcustomizer

Could also make the `requestForAuthEndpoint` function public so clients could use that to get the request without doing it themselves.

hamchapman:
Okay this is now in the state I'd like it to be in (bar potentially some small README updates).

@Noobish1 - happy with it as it is?

I'll push out a 2.0.0 release very soon if you are.

Ideally I'd wait for https://github.com/Carthage/Carthage/pull/1373 to get merged as then I can be lazy and make travis upload the built and zipped PusherSwift.framework.zip for Carthage users, but it's not a deal-breaker

Noobish1:
Looks good to me :+1:

PR #62, opened 2016-06-24
Feature/pass all params to authrequestcustomizer

hamchapman:


hamchapman:
Incorporated into #60

PR #63, opened 2016-06-28
Fixed option `attemptToReturnJSONObject`

dral3x:
That option allow us to receive a plain string as data but in reality it does not work.
The check was about having a value from the options and did not check for its real value.

This commit fix this.
Hope it will be merged soon 👍


hamchapman:
Thanks @dral3x - this might actually be fixed in #60 but I need to check.

If it's not, then I'll add it in there.

PR #65, opened 2016-06-29
Fix autoreconnect

dral3x:
After some debugging, I fixed the autoreconnect feature.
Now, the retry happens only if 
1- the flag is set to `true` in the options
2- the socket disconnection occurred because of an error

Hope you merge this asap in master. Thanks

### DIFF
--- a/Source/PusherChannel.swift
+++ b/Source/PusherChannel.swift
@@ -84,7 +84,7 @@ public class PusherChannel {
     */
     public func handleEvent(eventName: String, eventData: String) {
         if let eventHandlerArray = self.eventHandlers[eventName] {
-            if let _ = connection.options.attemptToReturnJSONObject {
+            if let attempt = connection.options.attemptToReturnJSONObject where attempt {
                 for eventHandler in eventHandlerArray {
                     eventHandler.callback(connection.getEventDataJSONFromString(eventData))
                 }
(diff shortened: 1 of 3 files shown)


hamchapman:
I'm not sure how I feel about the changes in the Starscream file. I'd like to be able to keep tracking the Starscream library without any custom code if possible.

Are those changes required?

dral3x:
Oh ok. Well, the problem is that Starscream generate an NSError on disconnect even if the connection is closed with `Normal` code. To me this is not correct but if we want to leave that library as it is, we need to filter out that error when it arrives.

hamchapman:
Yeah, I've just tried this myself and it does seem to be the case.

I think we can check if the error's `localizedDescription` is `"connection closed by server"` and the error code is 1000 (normal close code).

If that's the case then don't try and reconnect. How does that sound to you?

dral3x:
Done.
I will not check for the error message (makes no sense). I checked for the code value.

Source/PusherChannel.swift
```diff
@@ -84,14 +84,11 @@ public class PusherChannel {
...
-            if let _ = connection.options.attemptToReturnJSONObject {
-                for eventHandler in eventHandlerArray {
-                    eventHandler.callback(connection.getEventDataJSONFromString(eventData))
-                }
-            } else {
-                for eventHandler in eventHandlerArray {
-                    eventHandler.callback(eventData)
-                }
+            
+            let jsonize = connection.options.attemptToReturnJSONObject ?? false
```
hamchapman: Why add the `?? false` ?

hamchapman: Oh sorry it's because it's not based on #60

dral3x:
Thanks man :)

PR #66, opened 2016-07-04
More robust reconnect

hamchapman:
Handles reconnection when Pusher server closes the websocket connection

@psycotica0 WDYT?


psycotica0:
I haven't had a chance to look too closely at the interactions with the stuff I had, but as a user of the library, I might want 3 more things from this (sorry) 

If it doesn't just attempt to reconnect forever, then I'd probably want a connection state to differentiate between reconnecting (or delaying before attempting to reconnect) and disconnected for good and never going to try again. 
With this logic I think the connection state would switch to disconnected in either state. 

It's also the case that someone might want their app to try again forever.
The library user can emulate that now by just picking a really big number, but it might be more interesting to make it optional or something, in which case it just goes forever. 

Final thing, in the infinite or even large number cases the exponential back off might get big. Which is its job, but as a developer of something that consumes pusher as what the app primarily does, I would rather eat up CPU cycles trying to reconnect than wait 10 minutes between attempts. 
So, I think I would prefer a sort of configurable maximum retry delay, so it does the exponential thing up to this point then it stays here. 

In my use case I have an app that doesn't do anything unless pusher is connected, so I'd probably set retry to infinite and max time to 5 or 10 seconds, because anytime it's not connected is time it's doing nothing. 

If, though, I did want to do something, I would want to know when to wait and let it do its thing, and when it's actually dead, which was the first point I made. 

I'll look at the technical side later today?

hamchapman:
All very valid points. Thanks! I'll make some changes to reflect the suggestions now. Really keen to get 2.0.0 out soon!

hamchapman:
I've made updates to accommodate your points about a maximum time between reconnect attempts and having a maximum total number of reconnect attempts.

I've not made any connection state changes yet. Ideally I'd like some input on how you think they should work.

I was thinking that it could be something like this.

| App state  | Connection state |
| --- | --- |
| Starting up  | `.Disconnected`  |
| `pusher.connect()` called  | `.Connecting`  |
| Connection success message received  | `.Connected`  |
| Pusher server closes websocket  | `.Disconnected`  |
| Reconnect logic kicks in  | `.Reconnecting`  |
| Connection attempt about to be made  | `.Connecting`  |
| Reconnection attempt fails  | `.Reconnecting`  |
| Connection attempt no. 2 about to be made  | `.Connecting`  |
| Connection success message received  | `.Connected`  |

So it would just mean adding one new connection state: `Reconnecting`

I'm not sure how that would look in code yet, but that's what I was thinking of trying. Almost needs a proper state machine.

Source/PusherConnection.swift
```diff
@@ -653,6 +662,28 @@ public class PusherConnection {
                 )
             }
     }
+
+    /**
+        Attempt to reconnect triggered by a disconnection
+    */
+    @objc internal func attemptReconnect() {
+        if connectionState != .Connected {
+            if (reconnectAttemptsMax == nil || (reconnectAttemptsMax != nil && reconnectAttempts < reconnectAttemptsMax)) {
```
psycotica0: If the second side of your `||` is evaluated, it's because the first side is false, so in
```swift
reconnectAttemptsMax == nil || (reconnectAttemptsMax != nil && reconnectAttempts < reconnectAttemptsMax)
```
You don't have to check `!= nil`, it's implied by the or.

```swift
reconnectAttemptsMax == nil || reconnectAttempts < reconnectAttemptsMax
```

Source/PusherConnection.swift
```diff
@@ -19,6 +19,10 @@ public class PusherConnection {
     public var userDataFetcher: (() -> PusherUserData)?
     public var debugLogger: ((String) -> ())?
     public weak var stateChangeDelegate: ConnectionStateChangeDelegate?
+    public var reconnectAttemptsMax: Int? = 6
+    public var reconnectAttempts: Int = 0
+    public var maxReconnectGapInSeconds: Int? = nil
```
psycotica0: You invented `maxReconnectGapInSeconds`, and the only place you're using it it's being cast to a `Double`.

Is there a reason it can't just be defined as `Double`?
It's not abnormal in iOS to use a double to represent seconds (https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_DataTypes/index.html#//apple_ref/c/tdef/NSTimeInterval)

Source/PusherConnection.swift
```diff
@@ -653,6 +662,28 @@ public class PusherConnection {
...
+                let reconnectInterval = Double(reconnectAttempts * reconnectAttempts) * 2.0
+                let timeInterval = maxReconnectGapInSeconds != nil ? max(reconnectInterval, Double(maxReconnectGapInSeconds!))
+                                                                   : reconnectInterval
+                reconnectTimer = NSTimer.scheduledTimerWithTimeInterval(
+                    timeInterval,
+                    target: self,
+                    selector: #selector(attemptReconnect),
+                    userInfo: nil,
+                    repeats: false
+                )
```
psycotica0: Is it bad to call `.connect()` on an already connecting websocket?

The call to `connect` above looks asynchronous, which means that if we take 2 seconds between attempt 1 and 2, but the websocket is slow and was going to take 3 seconds to connect, then we will call `connect`, wait 2 seconds, then call it again but the connection hasn't necessarily failed yet?

I don't know if that's a real concern or not, just bringing it up.

hamchapman: An excellent point. Looks like we do need to be careful not to call `connect()` multiple times on the same socket. Not sure on the best solution: 

* add a check for the connection state being `.Connecting` and if it is, return, much like the library already does if already connected 
* setup an operation queue for connection-based operations (namely `connect()` and `disconnect()`)

WDYT?

hamchapman: In fact having the queue doesn't even guarantee that you won't end up calling `connect()` when the state is already `.Connecting`

psycotica0:
Ok, before we get to connection state, a couple notes about reconnect.

First, I think this NSTimer stuff completely removes the need for the NSOperation stuff.

With both, when the socket goes down we're queueing up a block to run an operation in 1 second, then calling attemptReconnect which both calls connect immediately, and queues up another operation to try again 2 seconds from now.

The NSOperation stuff isn't useful in here, as the Timer has it covered.

But, there are a couple things I did with the NSOperation that aren't being similarly addressed here, specifically with regards to Reachability.

In this implementation if I have my attempts set to 6 and I turn the network off, it looks like it will fail, then attempt to reconnect 6 times while the network is unreachable, then it will exhaust that and stop trying. Then we will turn the network back on and it will reconnect, because that's what it does when reachability comes up.

Probably not intended.
I don't know what you want to do with the interplay between reconnect and reachability.

You could ignore reachability entirely and just do reconnect, like this, and failure is failure, regardless of why, and after 6 we're done, or you could do something kinda like what the NSOperation stuff did, which is that if the network is unreachable, we don't attempt.
In that case, you'd cancel the timer in reachability down (removing the NSOperation), and call `attemptReconnect` in reachability up. (Or keep calling `connect`, like now, and let failure handle the reconnect like it already is)

In that case you might have reachability go down, it would reattempt immediately, that would fail and it would mark the number of attempts up to 1, then it would enqueue another attempt 2 seconds out.
Then, reachabiity goes down, and we stop the timer essentially "pausing" reconnection.

Then when reachability comes back up we resume (already at 1 attempt) and if that fails, we mark it up to 2 and wait 8 seconds before trying again, like normal.

I'm out of time for now, but those are my first thoughts. I'll talk more about statuses later.

psycotica0:
In previous versions, both before and after the reconnect stuff I added, we didn't need a reconnecting status, because it was always reconnecting. 

If you had auto reconnect set, disconnected implied reconnecting. 

The need for a new status now is to differentiate, to the host application, between "I have disconnected, but don't worry about it, I've got it covered" and "I've disconnected and you had better take action or I'll stay that way"

With this in mind, I think I'd prefer if auto reconnect is off, as soon as it disconnects it goes to disconnected, but if auto reconnect is on it goes directly to reconnecting and stays there until either it is connected or it runs out of attempts, in which case it goes to disconnected. 

That way I can differentiate between reconnecting, in which case I might show a little badge telling the user to be patient, and disconnected where I might pop up a box saying "hey, we're having trouble reconnecting, click here to retry" or something. 

If it went to disconnected before reconnecting then I'll pop the box up too early. For example.

hamchapman:
Again, all sounds very sensible to me. I'll whip something up that hopefully encompasses all of this.

Source/PusherConnection.swift
psycotica0: Well, what we could maybe do is recast the delay, not as one between connection attempts, but one between failure and reattempt.

In that case we don't need to have `attemptReconnect` enqueue itself, we only enqueue in `websocketDidDisconnect` with the computed delay.

(Still cancelling in the successful connection case)

So, if we attempt to reconnect at time 0, and it takes 3 seconds for the websocket to give up, and our delay would be 2 seconds, it takes 5 seconds total for the next connection attempt to begin.

The most important thing here is to be careful that all calls to `connect` either do nothing if there's already an enqueued retry attempt that will do it instead, or cancel the enqueued attempt and try immediately.

hamchapman:
Okay I've pushed some changes that make the logic easier to follow (in my opinion at least!) and, from my testing, makes reconnection more reliable.

The last thing I want to do is change the connection states so that there's an extra one called something like `.ReconnectingWhenNetworkBecomesReachable`. Bit of a mouthful, but it gets the intent across

Source/PusherWebsocketDelegate.swift
```diff
@@ -31,36 +30,71 @@ extension PusherConnection: WebSocketDelegate {
...
+        // Attempt reconnect if possible
+
+        guard self.options.autoReconnect else {
+            return updateConnectionState(.Disconnected)
+        }
+
+        guard reconnectAttemptsMax != nil && reconnectAttempts < reconnectAttemptsMax! else {
+            self.debugLogger?("[PUSHER DEBUG] Max reconnect attempts reached")
+            return updateConnectionState(.Disconnected)
+        }
```
psycotica0: I think this is the same as:

```swift
guard let reconnectAttemptsMax = reconnectAttemptsMax where reconnectAttempts < reconnectAttemptsMax else {
    [disconnect]
}
```

which makes it clear that providing a `nil` to `reconnectAttemptsMax` is considered the same as setting `autoReconnect` to `false`, not providing an unlimited number of attempts.

I think you want `reconnectAttemptsMax == nil || [...]`

Source/PusherWebsocketDelegate.swift
```diff
@@ -31,36 +30,71 @@ extension PusherConnection: WebSocketDelegate {
...
+
+        attemptReconnect()
+    }
+
+    /**
+        Attempt to reconnect triggered by a disconnection
+    */
+    internal func attemptReconnect() {
+        if connectionState != .Connected {
+            if (reconnectAttemptsMax == nil || reconnectAttempts < reconnectAttemptsMax!) {
```
psycotica0: I like what the `guard` syntax did to the above function, you could probably do that here too.

psycotica0:
I like this much better, yeah! (besides those couple things)

psycotica0:
So, I guess the intention of `.ReconnectingWhenNetworkBecomesReachable` is to go in the part of `websocketDidDisconnect` where we check reachability and are currently calling that `.Disconnected`?

Makes sense.
So, this is, I would assume, to potentially signify to the user that their network is really down, but if we see it come up we'll reconnect. I could see that being a state the application might want to know about, sure.

Better than disconnect!

Source/PusherWebsocketDelegate.swift
hamchapman: Maybe I'm misunderstanding, but I _think_ it's correct as it is.

If `reconnectAttemptsMax` is `nil` then I'm assuming that the developer wants the library to keep on attempting to reconnect forever, assuming that they've not set `autoReconnect` to `false`.

So I want to check that they've _not_ set `reconnectAttemptsMax` to `nil` (otherwise I can skip checking how many attempts have been made) and if they haven't, then also check if the maximum number of reconnect attempts have been made.

I think that's what the code does at the moment. Let me know if I've got that wrong though!

psycotica0: Guard syntax, unless I'm mistaken, says "don't move past this block unless this is true", which is how you've been using it throughout the code.

So when you say `guard self.options.autoReconnect else { Y }`, you're saying "every line beneath this one I should be able to rely on `self.options.autoReconnect` being true.

Or, put another way, it does this transformation (which you just did in your latest commit, which looks good)

```swift
guard X else {
    Y
}
Z
```
into
```swift
if X {
    Z
} else {
    Y
}
```

It makes some requirements on `Y` that it must be an exit, and in exchange we get to keep our nesting down, because `Z` doesn't have to be wrapped up in the `if`'s braces.

So, this statement is guarding `reconnectAttemptsMax != nil && reconnectAttempts < reconnectAttemptsMax!`, which means everywhere below this line that above statement must be true.

It's asserting that `reconnectAttemptsMax` is not nil. So, if `reconnectAttemptsMax` is nil, then we should exit early (in this case setting the state to `Disconnected` on the way).

If you switch it to `||` like you have on line 81 of Source/PusherWebsocketDelegate.swift then you'll be saying "To continue attempting to reconnect either the configuration must be `nil`, or if it exists it must be greater than the number of attempts I've made so far`, which is the semantic I believe you're going for.

I could be wrong, though. In this case it's pretty easy to test and verify what it's actually doing.
You can put a log in both places, set `reconnectAttemptsMax` to `nil`, and just see which way it goes.

hamchapman:
You were definitely right about that guard condition. Not sure why I seemed to grok how to use it properly in all cases apart from that one!! Thanks anyway.

And I've added the `ReconnectingWhenNetworkBecomesReachable` state. And some more docs about how it's all working.

Going to merge later today and release 2.0.0 unless you've got any further comments 🎉

Source/PusherWebsocketDelegate.swift
```diff
@@ -31,36 +30,75 @@ extension PusherConnection: WebSocketDelegate {
...
+        guard reconnectAttemptsMax == nil || reconnectAttempts < reconnectAttemptsMax! else {
+            return
+        }
+
+        let reconnectInterval = Double(reconnectAttempts * reconnectAttempts)
+
+        let timeInterval = maxReconnectGapInSeconds != nil ? min(reconnectInterval, maxReconnectGapInSeconds!)
+                                                           : reconnectInterval
+
+        self.debugLogger?("[PUSHER DEBUG] Waiting \(timeInterval) seconds before attempting to reconnect (attempt \(reconnectAttempts + 1) of \(reconnectAttemptsMax!))")
```
psycotica0: Pretty sure this will crash if `reconnectAttemptsMax` is `nil`, since you're forcing it here, but we know it might not exist.
I'm not sure if `debugLogger` is required for it to crash, since Swift might put the string together as an argument even if it passes it to a function which is ignored. I don't know if it does or not.

If it does, then this will crash every time `reconnectAttemptsMax` is `nil`, otherwise it will only crash when it's `nil` and you have `debugLogger` set.

hamchapman: Ah of course. Let that slip through. Thanks. I'll publish a 2.0.1 release later today with a fix.

hamchapman: Publish 2.0.1 - thanks again for the spot 👍

PR #69, opened 2016-07-20
Update pod to ios8 version

ZalyalovIldar:
You'r deployment target set to iOS 8, but in your Pod file was iOS 9, 'cause of this I can't install 2.0.1 version.


hamchapman:
This shouldn't be making a difference. The change you made is in relation to the tests target, which uses pods for getting Nimble and Quick into the build.

You can see that in the podspec file the iOS version is set to 8.0: https://github.com/pusher/pusher-websocket-swift/blob/master/PusherSwift.podspec#L14

PR #78, opened 2016-08-15
allow authRequestBuilder's to fail when creating requests for authorization

Noobish1:
By returning a non-optional we don't allow conforming objects to fail to create requests, which is possible.

### DIFF
--- a/Source/PusherClientOptions.swift
+++ b/Source/PusherClientOptions.swift
@@ -19,7 +19,7 @@ public enum PusherHost {
 }
 
 public protocol AuthRequestBuilderProtocol {
-    func requestFor(socketID: String, channel: PusherChannel) -> NSMutableURLRequest
+    func requestFor(socketID: String, channel: PusherChannel) -> NSMutableURLRequest?
 }
 
 public enum AuthMethod {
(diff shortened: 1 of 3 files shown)


hamchapman:
Seems reasonable. 

semver says this should be a bump to 3.0.0, so I guess that's what's gonna happen.

While we're here, do you think it's worth adding an optional callback that will be called when a subscription fails because authorisation fails?

Noobish1:
funny you mention that as I'm in need of that exact thing. It'd either be a callback when auth fails or one callback for when the we connect or fail to connect.

hamchapman:
Yeah, it's not super clear (or maybe not clear _at all_) but you can already bind the channel to an event `pusher_internal:subscription_succeeded` and provide a function there to be run when the subscription is successful.

hamchapman:
In fact it looks like it gets re-emitted as just `pusher:subscription_succeeded`

Noobish1:
Yea I had a dig through, that global callback stuff isn't very obvious. Matching on a string is yuck as well.

hamchapman:
Having a specific callback is certainly nicer :)

PR #90, opened 2016-09-19
Remove tryFlushOutbox and replace with a TaskQueue

hamchapman:
(Also made use of connection delegate debugLog function in NativePusher)

### DIFF
--- a/Source/TaskQueue.swift
+++ b/Source/TaskQueue.swift
@@ -1,0 +1,217 @@
+/**
+    The MIT License (MIT)
+
+    Copyright (c) 2014 Marin Todorov
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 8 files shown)


hamchapman:
@jameshfisher can you take a look at this please?

Source/TaskQueue.swift
```diff
@@ -0,0 +1,217 @@
+/**
```
jameshfisher: Is this the way dependencies are managed in Swift?

hamchapman: They don't have to be but I've found that it's easier to do it this way. 

Although you have reminded me that I need to make the access level `internal` rather than `open`. Will do that in a sec

jameshfisher:
All makes sense - can't say it works but it makes sense!

Source/TaskQueue.swift
jameshfisher: Oh I also see you've altered it! Okeydoke

jameshfisher:
@hamchapman GG

PR #97, opened 2016-10-06
Changed a few references to NSURLRequest/NSMutableURLRequest to URLRequest

Noobish1:


hamchapman:
I think I had this originally when working on the Swift 3 branch but then I believe I needed to keep `NSMutableRequest` for the Obj-C compat. I'll have a check though.

Do you need the type to be `URLRequest` for your use case?

Noobish1:
We do yea

hamchapman:
Just been testing this locally and it does in fact seem to work fine with the Obj-C stuff, so this will likely go into a 4.0.1 release 👍

hamchapman:
Closed in favour of #107

PR #98, opened 2016-10-07
Adding code to suppress default warning when unused method result in …

nickjf89:
…Swift 3


hamchapman:
I think I might instead mark the bind and functions with `@discardableResult` so that the warnings don't appear

nickjf89:
Sounds good to me 👍🏻

—

Nick Farrant
07720635147

hamchapman:
Added in 0030fa43

Issue #101, opened 2016-10-20
Memory leaks in manager.connect()


bimusiek:
Hey guys,
we are investigating memory leaks with PusherManager and we found that method `connect` leaks quite a lot for us (~700kb) while calling it 3 times.

We have in our one instance init class

``` swift
 let options = PusherClientOptions(
            authMethod: AuthMethod.authRequestBuilder(authRequestBuilder: AuthRequestBuilder(delegate: self))
        )
        manager = Pusher(
            key: config.PUSHER_KEY,
            options: options
        )
```

so `Pusher` is always one instance.

Then, when someone change user in our app we call

``` swift

        self.privateChannel?.unbindAll()
        self.privateChannel = nil

        self.manager.unbindAll()
        self.manager.disconnect()

        self.manager.connect()

        let channelName = "private-user-\(userId)"
        DLOG("Subscribing to private channel named: \(channelName)")
        privateChannel = self.manager.subscribe(channelName)
        _ = privateChannel?.bind(eventName: "update", callback: { [weak self] (response) in
            DLOG("Pusher update: \(response)")
            if let data = response as? [String:AnyObject] {
                let type = data["type"] as? String
                let eventType = PusherEventEntity.fromString(type)

                self?.invokeDelegate({ (delegate) -> () in
                    delegate.updateEvent(eventType, data: data as AnyObject)
                })
            }
        })
```

<img width="1032" alt="screenshot 2016-10-20 18 11 54" src="https://cloud.githubusercontent.com/assets/1025588/19568154/bca46efc-96f0-11e6-9213-85aef3b2d1bd.png">

Any ideas?

bimusiek:
If someone had the same issue, here is solution.

You don't have to disconnect and connect when you switch user. Pusher does authentication when subscribing, so we connect once now and then we unsubscribe and subscribe again. :) No more leaks.

hamchapman:
Interesting! 

Thanks for writing this up. I'll hopefully get some time soon to take a look into this.

anlaital:
This was also causing us considerable pain so I fixed this quickly. This could use some refactoring, but I suppose that's something up for discussion. :)

There is now a pull request for this https://github.com/pusher/pusher-websocket-swift/pull/105

anlaital:
[...]

PR #105, opened 2016-11-10
Quick fix for a couple of memory leaks.

anlaital:
Fixes memory leaks as described in https://github.com/pusher/pusher-websocket-swift/issues/101

Note that this is just a quick fix for the issue that's affecting us and causing pretty big memory leaks. I think that a better fix would be to refactor how `PusherChannel` uses the facilities provided by `PusherConnection` as now it's just given the entire class instance to use.


hamchapman:
Good point about the channels not needing the whole connection object. I'll take a look at that today.

PR #110, opened 2016-11-21
Update link to macOS example

BasThomas:
The `AppDelegate` contains the example, not the `ViewController`.


hamchapman:
Good point! The macOS example does currently only show the native push code example. Thanks!

PR #112, opened 2016-11-29
Fixing init calls in readme and obj-c example

nickjf89:


hamchapman:
Turns out both are valid so going to leave it as is

Issue #117, opened 2017-01-13
Channel name is not urlencoded before sending to authEndpoint


Taik:
The current library does not URL encode the channel name before sending to `authEndpoint`. 

The HMAC signatures will be different depending on if the channel name is encoded or not; it seems like the signature returned from pusher is the urlencoded version.

Bellow are a few sample POST body, sent by different libraries.

- pusher/pusher-websocket-swift:
```json
{"level":"info","ts":1484339706.0635204,"msg":"pusher.go:104: Debug","payload":"socket_id=208841.2081733&channel_name=private-reservations-for-venue@venue_id=399ccd2d-3f4a-43c9-803c-9e4b6bdf0f16;date=2017-01-13"}
```
- pusher/pusher-js:
```json
{"level":"info","ts":1484340851.4951386,"msg":"pusher.go:104: Debug","payload":"socket_id=208909.2093154&channel_name=private-reservations-for-venue%40venue_id%3D399ccd2d-3f4a-43c9-803c-9e4b6bdf0f16%3Bdate%3D2017-01-13"}
```

The same channel names are used; but the request via the swift library returns an invalid signature error:
```
Invalid signature: Expected HMAC SHA256 hex digest of 208879.2102145:private-reservations-for-venue@venue_id=399ccd2d-3f4a-43c9-803c-9e4b6bdf0f16;date=2017-01-13, but got f12759368cf888c8f5f63276587df9c9940bdce61a477013ab66d122a76f0f8f
```

hamchapman:
Thanks for spotting this!

I'll look at getting this fixed this coming week, or feel free to make a PR if you're happy to 👍

Taik:
Closing as #119 was merged :)

hamchapman:
Merged the provided pull request and added a test and a small tweak

PR #119, opened 2017-01-18
Encode channel name in request body for authEndpoint

natechiger:
This is to fix the following issue: https://github.com/pusher/pusher-websocket-swift/issues/117


hamchapman:
Thanks for this! Going to merge and make a couple of tiny tweaks and add a test 👍

Issue #115, opened 2016-12-22
Starscream crash


hamchapman:
See #113.

Summary:

```
Crashed: com.vluxe.starscream.websocket
0  libobjc.A.dylib                0x18a4faf3c objc_msgSend + 28
1  CoreFoundation                 0x18ba4d5ec _inputStreamCallbackFunc + 56
2  CoreFoundation                 0x18b9f8258 _signalEventSync + 212
3  CoreFoundation                 0x18ba6963c ___signalEventQueue_block_invoke + 24
4  libdispatch.dylib              0x18a939200 _dispatch_call_block_and_release + 24
5  libdispatch.dylib              0x18a9391c0 _dispatch_client_callout + 16
6  libdispatch.dylib              0x18a947444 _dispatch_queue_serial_drain + 928
7  libdispatch.dylib              0x18a93c9a8 _dispatch_queue_invoke + 652
8  libdispatch.dylib              0x18a947940 _dispatch_queue_override_invoke + 360
9  libdispatch.dylib              0x18a94938c _dispatch_root_queue_drain + 572
10 libdispatch.dylib              0x18a9490ec _dispatch_worker_thread3 + 124
11 libsystem_pthread.dylib        0x18ab422b8 _pthread_wqthread + 1288
12 libsystem_pthread.dylib        0x18ab41da4 start_wqthread + 4
```

Also need to rename the embedded Starscream queue, for example, to `com.starscream.websocket.pusher`.

zkyronex:
@hamchapman Really need to know if this issue has been looked at. This bug will affect anyone's confidence to use the library in their app as a reliable communication option. I am currently looking into this as well.

Crashes from `com.apple.root.default-qos.overcommit` and not on the main thread.
Investigating according to this lead:
http://stackoverflow.com/questions/27948618/consistent-dispatch-queue-com-apple-root-default-qos-overcommit-crash

Any follow up on this would be greatly appreciated. Cheers!

hamchapman:
@zkyronex have you experienced the same issue?

I've tried to recreate it but have been unable to do so and so I'm really just waiting for more information.

I'm going to make a branch that updates to the latest version of Starscream on master and also change the queue name that the Pusher library uses for the underlying Starscream dispatch queue.

zkyronex:
[...]

PR #123, opened 2017-01-27
Reachability fix

QuimeraKoke:
Hi,
I fix the issue #115 as I explained in the thread.

* added async at whenReachable and whenReachable block
* let self retained in whenReachable and whenReachable block

### DIFF
--- a/Source/PusherConnection.swift
+++ b/Source/PusherConnection.swift
@@ -38,24 +38,18 @@ open class PusherConnection: NSObject {
 
     open lazy var reachability: Reachability? = {
         let reachability = Reachability.init()
-        reachability?.whenReachable = { [weak self] reachability in
-            guard self != nil else {
-                print("Your Pusher instance has probably become deallocated. See https://github.com/pusher/pusher-websocket-swift/issues/109 for more information")
-                return
-            }
-
-            self!.delegate?.debugLog?(message: "[PUSHER DEBUG] Network reachable")
-            if self!.connectionState == .disconnected || self!.connectionState == .reconnectingWhenNetworkBecomesReachable {
-                self!.attemptReconnect()
+        reachability?.whenReachable = { reachability in
+            DispatchQueue.main.async {
+                self.delegate?.debugLog?(message: "[PUSHER DEBUG] Network reachable")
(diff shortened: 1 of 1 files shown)


theMeatloaf:
I'm using this PR version in my production app and it has somehow caused my connections to increase...I believe it has caused a memory leak which is duplicating pusher objects and thus increasing the connection count.

hamchapman:
Thanks for the info @theMeatloaf. I think that's what I would have expected seeing as the `[weak self]` is removed by this PR. 

Can I ask why you were testing this PR? Were you also seeing the starscream crash?

theMeatloaf:
Yes I was seeing the starscream crash in high volumes, essentially making the lib unusable.  I will look into a better solution, but i'm not fully versed on the cause so we'll see what I can come up wtih

theMeatloaf:
isnt the solution to just keep the `[weak self]` isntead of `unowned` ?

theMeatloaf:
Yeah looking at the basic area of the crash, it appears that a `weak` reference to self is what is needed here.  @hamchapman If you think the current version of this file in the repo will fix the issues i'd be happy to try it out.

hamchapman:
Agreed that having a `weak` reference to `self` in the closure is what is needed. And that's what it currently has on master. I'm still just unsure as to how the rest of the changes in this PR would fix any kind of Staracream queue crash

theMeatloaf:
I'm not sure i fully understand the Starscream crash but it sounds like the memory leak generated by this PR had the side effect that it fixed it since the referenced objects that previously were getting deallocated have leaked and are no longer crashing...

The team at pusher keeps contacting our team about hitting our connection limit so i think i'm going to try and see if the current master branch fixes my issues... i'll let you know @hamchapman

hamchapman:
@theMeatloaf did you find that master fixed your issues?

theMeatloaf:
Yes it does appear to have fixed my issues.  I occasionally do see the "Your pusher instance has been deallocated" message in the log but it still continues to respond to messages

hamchapman:
Thanks for letting me know. I'm going to close this for now as I don't think it's the right way to fix any potential problems with that bit of the codebase.

PR #128, opened 2017-02-17
Update Starscream.swift

scartick:
fix problem with starscream for swift 2.3


hamchapman:
Updated in the 4.1.0 release

PR #129, opened 2017-02-20
Manual auth

hamchapman:
Updates the `subscribe` and `subscribeToPresenceChannel` functions to accept and optional `auth` parameter of type `PusherAuth`.

This `PusherAuth` object can be initialised with just an `auth (String)` value if the subscription is to a private channel, or both an `auth (String)` and `channelData (String)` pair of values if the subscription is to a presence channel.

These `auth` and `channelData` values are the values that you received if the json object created by a call to `pusher.authenticate(...)` in one of our various Ruby libraries.

Keep in mind that in order to generate a valid `auth` value for a subscription the `socketId` (i.e. the unique identifier for a web socket connection to the Pusher servers) must be present when the `auth` value is generated. As such, the likely flow for using this is something like this:

```swift
import UIKit
import PusherSwift

class ViewController: UIViewController, PusherDelegate {
    var pusher: Pusher! = nil

    override func viewDidLoad() {
        super.viewDidLoad()

        let optionsWithEndpoint = PusherClientOptions(
            authMethod: AuthMethod.endpoint(authEndpoint: "https://some.endpoint.for.non.manual.auth")
        )
        pusher = Pusher(key: "your-app-key", options: optionsWithEndpoint)
        pusher.delegate = self
        pusher.connect()
    }

    // PusherDelegate methods

    func changedConnectionState(from old: ConnectionState, to new: ConnectionState) {
        // print the old and new connection states
        print("old: \(old.stringValue()) -> new: \(new.stringValue())")

        // we know we are connected here as we have a socketId
        if let socketId = pusher.connection.socketId {
            // a dummy channel name value here, just as an example
            let channelName = "presence-test"

            getAuthData(channelName: channelName, socketId: socketId, completionHandler: { res in
                // the response from the server looks like: `{ "auth": "123123abcabc123", "channel_data": "{\"user_id\":16,\"user_info\":{\"time\":\"2017-02-20 14:54:36 +0000\"}}"}
                let pusherAuth = PusherAuth(auth: res["auth"]!, channelData: res["channel_data"]!)
                let _ = self.pusher.subscribe(channelName, auth: pusherAuth)
            })
        }
    }

    func getAuthData(channelName: String, socketId: String, completionHandler: @escaping ([String: String]) -> ()) {
        var request = URLRequest(url: URL(string: "https://yourendpointforcreatingauthvalues.com")!)
        request.httpMethod = "POST"
        request.httpBody = "socket_id=\(socketId)&channel_name=\(channelName)".data(using: String.Encoding.utf8)

        let task = URLSession.shared.dataTask(with: request, completionHandler: { data, response, sessionError in
            guard let data = data else {
                print("No data")
                return
            }

            guard let httpResponse = response as? HTTPURLResponse, (httpResponse.statusCode == 200 || httpResponse.statusCode == 201) else {
                let dataString = String(data: data, encoding: String.Encoding.utf8)
                print ("Error getting auth data for [\(channelName)]: \(dataString)")
                return
            }

            guard let jsonObject = try? JSONSerialization.jsonObject(with: data, options: []), let json = jsonObject as? [String: String] else {
                print("Error parsing response data")
                return
            }

            completionHandler(json)
        })

        task.resume()
    }
}
```

### DIFF
--- a/Source/ObjectiveC.swift
+++ b/Source/ObjectiveC.swift
@@ -11,8 +11,22 @@ public extension Pusher {
         return self.subscribe(channelName, onMemberAdded: nil, onMemberRemoved: nil)
     }
 
+    @objc public func subscribe(
+        channelName: String,
+        onMemberAdded: ((PusherPresenceChannelMember) -> ())? = nil,
+        onMemberRemoved: ((PusherPresenceChannelMember) -> ())? = nil) -> PusherChannel {
+            return self.subscribe(channelName, auth: nil, onMemberAdded: onMemberAdded, onMemberRemoved: onMemberRemoved)
+    }
+
     @objc public func subscribeToPresenceChannel(channelName: String) -> PusherPresenceChannel {
-        return self.subscribeToPresenceChannel(channelName: channelName, onMemberAdded: nil, onMemberRemoved: nil)
+        return self.subscribeToPresenceChannel(channelName: channelName, auth: nil, onMemberAdded: nil, onMemberRemoved: nil)
+    }
+
+    @objc public func subscribeToPresenceChannel(
+        channelName: String,
+        onMemberAdded: ((PusherPresenceChannelMember) -> ())? = nil,
+        onMemberRemoved: ((PusherPresenceChannelMember) -> ())? = nil) -> PusherPresenceChannel {
+            return self.subscribeToPresenceChannel(channelName: channelName, auth: nil, onMemberAdded: onMemberAdded, onMemberRemoved: onMemberRemoved)
     }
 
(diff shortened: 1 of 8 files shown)


pedromcunha:
Hey just pinging @hamchapman, any idea when this will get merged in?

hamchapman:
Have you managed to get a chance to test it? If so, was it all working as you'd expect?

Basically as soon as I've had someone other than myself test it out then I'll be happy to merge it 👍

hamchapman:
Did you get a chance to test this out @pedromcunha ?

PR #141, opened 2017-06-08
Updating @_specialized attributes to work with >=Swift 3.2

BrendanBuono:
Description of the pull request

This pull requests updates the usage of @_specialized to work with all versions of swift
...

Why is the change necessary?

Swift 3.2 and 4 changed the usage of @_specialized to allow for multiple generic constraints, which requires a swift macro to have both versions in code

____

CC @pusher/mobile

### DIFF
--- a/Source/CryptoSwiftHMACModule.swift
+++ b/Source/CryptoSwiftHMACModule.swift
@@ -34,7 +34,7 @@ extension Array {
 }
 
 extension Array {
-
+    
     /** split in chunks with given chunk size */
     func chunks(size chunksize: Int) -> Array<Array<Element>> {
         var words = Array<Array<Element>>()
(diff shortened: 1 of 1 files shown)


DanSessions:
Perhaps this could be merged in to its own branch, and not master?

hamchapman:
I think it would be better to have a branch that just works with Swift 3.2+ and drops support for older versions of Swift. That way the Swift version check can be dropped.

Also, there seems to be a lot of whitespace changes in the pull request. Please could you remove those changes from the pull request? It adds a lot of unnecessary noise to the diff. 

Thanks for the PR though 😁

BrendanBuono:
Yep, not a problem, I forgot to normalize the line endings. 

I contemplated having a pure 3.2 implementation but I wanted to maintain backwards compatibility for users, I'm cool with making the switch though. I'll update the PR later tonight

kielgillard:
My team is very interested in these fixes and looking forward to them merged and released so we can start building and running our app with Xcode 9.

sibinbk:
We are so keen with this update and start using it with Xcode 9.

johnrogers:
Do you guys have an update on when this will be merged? We're relying on Pusher in our projects and would like to start using Xcode 9. Any updates you could give us devs would be fantastic. 👍

BrendanBuono:
I will finish this in the AM with the fixes requested by @hamchampan

BrendanBuono:
Due to merge conflicts, I killed this and moved it to #144

PR #144, opened 2017-06-16
Updating @_specialized attributes to work with >=Swift 3.2

BrendanBuono:
Description of the pull request

This pull requests updates the usage of @_specialized to work with all versions of swift
...

Why is the change necessary?

Swift 3.2 and 4 changed the usage of @_specialized to allow for multiple generic constraints, which requires a swift macro to have both versions in code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode8
+osx_image: xcode9
 sudo: false
 env:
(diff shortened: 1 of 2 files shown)


BrendanBuono:
@hamchapman this should be good to go, let me know if you need anything else fixed

BrendanBuono:
FYI this will fail the CI as it's design to only work with swift-3.2 and above

PR #148, opened 2017-08-10
Unsubscribe all

dmorrow:
Convenience method to unsubscribe from all channels at once.

### Why is the change necessary?

In a multi-user environment, on signing out, it is useful to unsubscribe from all channels.
In testing builds, its even more useful, as users have different ids across environments.

----

CC @pusher/mobile

### DIFF
--- a/Source/PusherConnection.swift
+++ b/Source/PusherConnection.swift
@@ -181,7 +181,7 @@ open class PusherConnection: NSObject {
     */
     internal func unsubscribeAll() {
         for (_, channel) in channels.channels {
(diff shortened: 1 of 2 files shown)


hamchapman:
Thanks for the PR!

It would be awesome if you could add a test as well. If you're not up for it then we can add one 👍

dmorrow:
@hamchapman sorry - I don't see any appropriate examples of how you are testing other similar methods, so I'm going to leave the testing up to you

hamchapman:
Closing in favour of #164

PR #155, opened 2017-09-28
Disable Code Coverage

mattbanach:

Flip `CLANG_ENABLE_CODE_COVERAGE` to `NO`

Per Apple: “Do not submit apps with LLVM profiling instrumentation or
coverage collection enabled. Turn off LLVM profiling or code coverage,
rebuild your app and resubmit the app.”

CC @pusher/mobile


hamchapman:
Do you know if it need to be set to `NO` for both Debug and Release builds?

mattbanach:
As I understand it, you would only be affected for release builds.

hamchapman:
I've added this change but just for Debug builds in this commit: https://github.com/pusher/pusher-websocket-swift/commit/4c26917cdea8266bd993468b97db750c28a88934

Thanks for the PR!

PR #159, opened 2017-11-17
Fix warnings

lukabratos:

Fix warnings in the SDK.

To keep project well maintained ✨

----

CC @pusher/mobile

### DIFF
--- a/Sources/PusherConnection.swift
+++ b/Sources/PusherConnection.swift
@@ -581,7 +581,7 @@ public typealias PusherEventJSON = [String: AnyObject]
                 sendAuthorisationRequest(request: request, channel: channel)
                 return true
             case .authRequestBuilder(authRequestBuilder: let builder):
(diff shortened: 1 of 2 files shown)


Sources/PusherConnection.swift
```diff
@@ -581,7 +581,7 @@ public typealias PusherEventJSON = [String: AnyObject]
                 sendAuthorisationRequest(request: request, channel: channel)
                 return true
             case .authRequestBuilder(authRequestBuilder: let builder):
-                if let request = builder.requestFor?(socketID: socketId, channel: channel) {
+                if let request = builder.requestFor?(socketID: socketId, channelName: channel.name) {
```
hamchapman: This shouldn't be changed yet. I should have removed it with the release 5.0.0 but I forgot to do so. So now we need to wait until 6.0.0. See here: https://github.com/pusher/pusher-websocket-swift/blob/master/Sources/AuthRequestBuilderProtocol.swift#L5

PR #160, opened 2017-11-20
Implement set subscriptions functionality

lukabratos:

Users can now subscribe the client to all interests in the provided set as it is described [here](https://pusher.com/docs/push_notifications/reference/client_api#put-v1clientsclientidinterests).

To add additional functionality in the SDK.

----

CC @pusher/mobile

### DIFF
--- a/Sources/NativePusher.swift
+++ b/Sources/NativePusher.swift
@@ -166,6 +166,14 @@ import Foundation
         addSubscriptionChangeToTaskQueue(interestName: interestName, change: .unsubscribe)
     }
 
+    open func setSubscriptions(interests: Array<String>) {
+        requestQueue.tasks += {_, next in
+            self.modifySubscription(interest: interests, change: .setSubscriptions, successCallback: next)
+        }
+
+        requestQueue.run()
+    }
+
     /**
         Adds subscribe / unsubscribe tasts to task queue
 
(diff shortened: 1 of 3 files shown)


Sources/NativePusher.swift
```diff
@@ -187,29 +195,43 @@ import Foundation
...
         - parameter pusherAppKey: The app key for the Pusher app
         - parameter clientId:     The clientId returned by Pusher's server
-        - parameter interest:     The name of the interest to be subscribed to /
+        - parameter interest:     The name(s) of the interest(s) to be subscribed to /
                                   unsunscribed from
         - parameter change:       Whether to subscribe or unsubscribe
         - parameter callback:     Callback to be called upon success
     */
-    private func modifySubscription(interest: String, change: SubscriptionChange, successCallback: @escaping (Any?) -> Void) {
+    private func modifySubscription(interest: Any, change: SubscriptionChange, successCallback: @escaping (Any?) -> Void) {
```
hamchapman: I don't think we should change the type of `interest` to be `Any` here. There should be a separate function that handles the case for `interests` where it's `[String]`.

You will need to refactor `modifySubscription` a bit, probably into two (or more) separate functions, so that you can still reuse some of the logic in there for handling errors etc

Losing the type safety and relying on casting isn't what we should default to unless we have to.

lukabratos: I completely agree with you, @hamchapman. It's not the most elegant solution but it was the best compromise to do based on the discussion with @luismfonseca and state of the current implementation of the `NativePusher` class.

I'm more than happy to refactor and submit a new PR but I'll need to allocate more time for this task 👍

PR #161, opened 2017-11-20
Implement setSubscriptions

lukabratos:

Users can now subscribe the client to all interests in the provided set as it is described [here](https://pusher.com/docs/push_notifications/reference/client_api#put-v1clientsclientidinterests).

To add additional functionality in the SDK.

----

CC @pusher/mobile

### DIFF
--- a/Sources/NativePusher.swift
+++ b/Sources/NativePusher.swift
@@ -162,7 +162,14 @@ import Foundation
      - parameter interests: the name of the interests you want to subscribe to
      */
     open func setSubscriptions(interests: Array<String>) {
-        addSubscriptionChangeToTaskQueue(interests: interests, change: .setSubscriptions)
+        requestQueue.tasks += { _, next in
+            self.modifySubscription(
+                interests: interests,
+                successCallback: next
+            )
+        }
+        
+        requestQueue.run()
     }
 
     /**
@@ -188,17 +195,6 @@ import Foundation
             self.modifySubscription(
                 interest: interestName,
                 change: change,
-                successCallback: next
-            )
-        }
-
-        requestQueue.run()
-    }
-
-    private func addSubscriptionChangeToTaskQueue(interests: Array<String>, change: SubscriptionChange) {
-        requestQueue.tasks += { _, next in
-            self.modifySubscription(
-                interests: interests,
                 successCallback: next
             )
         }
--- a/Sources/NativePusher.swift
+++ b/Sources/NativePusher.swift
@@ -227,7 +227,6 @@ import Foundation
     /**
      Makes a PUT request for given interests
      - parameter interests:    The name of the interests to be subscribed to
-     - parameter change:       Whether to subscribe or unsubscribe
      - parameter callback:     Callback to be called upon success
      */
     private func modifySubscription(interests: Array<String>, successCallback: @escaping (Any?) -> Void) {
--- a/Sources/NativePusher.swift
+++ b/Sources/NativePusher.swift
@@ -163,7 +163,7 @@ import Foundation
      */
     open func setSubscriptions(interests: Array<String>) {
         requestQueue.tasks += { _, next in
-            self.modifySubscription(
+            self.replaceSubscriptionSet(
                 interests: interests,
                 successCallback: next
             )
@@ -192,7 +192,7 @@ import Foundation
     */
     private func addSubscriptionChangeToTaskQueue(interestName: String, change: SubscriptionChange) {
         requestQueue.tasks += { _, next in
-            self.modifySubscription(
+            self.subscribeOrUnsubscribeInterest(
                 interest: interestName,
                 change: change,
                 successCallback: next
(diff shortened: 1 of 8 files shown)


Sources/NativePusher.swift
```diff
@@ -186,91 +195,140 @@ import Foundation
...
-        try! request.httpBody = JSONSerialization.data(withJSONObject: params, options: [])
+        let url = "\(CLIENT_API_V1_ENDPOINT)/clients/\(clientId)/interests/"
+        let params: [String: Any] = ["app_key": pusherAppKey, "interests": interests]
+        let request = self.setRequest(url: url, params: params, change: change)
+        self.modifySubscription(interests: interests, request: request, change: change, successCallback: successCallback)
+    }
 
-        request.addValue("application/json", forHTTPHeaderField: "Content-Type")
-        request.addValue(LIBRARY_NAME_AND_VERSION, forHTTPHeaderField: "X-Pusher-Library")
+    private func modifySubscription(interests: Array<String>, request: URLRequest, change: SubscriptionChange, successCallback: @escaping (Any?) -> Void) {
```
jameshfisher: Does this function use the `change` param? I think if this overloaded version of `modifySubscription` is called, we assume that `change == setSubscriptions`

Sources/NativePusher.swift
```diff
@@ -186,91 +195,140 @@ import Foundation
         requestQueue.run()
     }
 
+    private func addSubscriptionChangeToTaskQueue(interests: Array<String>, change: SubscriptionChange) {
```
jameshfisher: This function is only called from one place, in `setSubscriptions(...)`? I think you can inline this function there

Sources/NativePusher.swift
```diff
@@ -188,89 +204,126 @@ import Foundation
...
+        return request
+    }
 }
 
 internal enum SubscriptionChange {
     case subscribe
+    case setSubscriptions
     case unsubscribe
 
     internal func stringValue() -> String {
```
hamchapman: I think you can remove this and change the enum to be defined as `internal enum SubscriptionChange: String` and then just call `.rawValue` (or something like that) on the enum values in the code, as opposed to having to call `.stringValue()` and defining values as below

Sources/NativePusher.swift
```diff
@@ -188,89 +204,126 @@ import Foundation
...
+        self.modifySubscription(interests: [interest], request: request, change: change, successCallback: successCallback)
+    }
 
-        let url = "\(CLIENT_API_V1_ENDPOINT)/clients/\(clientId!)/interests/\(interest)"
-        var request = URLRequest(url: URL(string: url)!)
-        request.httpMethod = change.httpMethod()
+    /**
+     Makes a PUT request for given interests
+     - parameter interests:    The name of the interests to be subscribed to
+     - parameter change:       Whether to subscribe or unsubscribe
```
hamchapman: `change` isn't present

Sources/NativePusher.swift
lukabratos: It's called by `subscribe` and `unsubscribe` where we pass `.subscribe` and `.unsubscribe`.

lukabratos:
PR updated! Thanks for the feedback guys 👍

Sources/NativePusher.swift
jameshfisher: We now have three different overloaded versions of `modifySubscription`:

```
modifySubscription(String, SubscriptionChange, successCallback)
modifySubscription(Array<String>, successCallback)
modifySubscription(Array<String>, URLRequest, SubscriptionChange, successCallback)
```

Why are we using overloading here? I think it makes it unclear what's being called. Instead we should use distinct function names for them:

```
subscribeOrUnsubscribeInterest(String, SubscriptionChange, successCallback)
replaceSubscriptionSet(Array<String>, successCallback)
doURLRequest(Array<String>, URLRequest, SubscriptionChange, successCallback)
```

Sources/NativePusher.swift
```diff
@@ -242,10 +242,10 @@ import Foundation
         let url = "\(CLIENT_API_V1_ENDPOINT)/clients/\(clientId)/interests/"
         let params: [String: Any] = ["app_key": pusherAppKey, "interests": interests]
         let request = self.setRequest(url: url, params: params, change: .setSubscriptions)
-        self.modifySubscription(interests: interests, request: request, change: .setSubscriptions, successCallback: successCallback)
+        self.doURLRequest(interests: interests, request: request, change: .setSubscriptions, successCallback: successCallback)
     }
 
-    private func modifySubscription(interests: Array<String>, request: URLRequest, change: SubscriptionChange, successCallback: @escaping (Any?) -> Void) {
+    private func doURLRequest(interests: Array<String>, request: URLRequest, change: SubscriptionChange, successCallback: @escaping (Any?) -> Void) {
```
lukabratos: I'm thinking if maybe `performURLRequest` would be a better name?

jameshfisher: also fine, sounds the same to me

lukabratos: Ok, I'll leave it as it is.

PR #162, opened 2017-11-30
Made Authorizer completion handler @escaping

BrandonShega:

Add `@escaping` to the parameter for the `Authorizer` protocol.

To allow the completion handler to be called from within a closure.

----

CC @pusher/mobile

### DIFF
--- a/Sources/Authorizer.swift
+++ b/Sources/Authorizer.swift
@@ -1,5 +1,6 @@
 import Foundation
 
 @objc public protocol Authorizer {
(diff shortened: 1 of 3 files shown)


DivineDominion:
Looks better :) However, the old tests don't exercise asynchronous callbacks: the closure is always executed synchronously. Not saying it's your job to fix that in this PR, but someone should keep that on their radar.

hamchapman:
Looks good to me.

@DivineDominion what do you mean when you say that the tests don't exercise asynchronous callbacks? Can you point me to the test(s) you're referring to?

hamchapman:
Replaced by #165

PR #181, opened 2018-04-05
Use Starscream instead of StarscreamFork

hamchapman:

See title

----

CC @pusher/mobile


andrewlord1990:
Do we have any ideas when this change might get released?

Currently, we are unable to update to Pusher Swift 6.0.0, due to a build error on CI because of StartScream, which has been fixed in their latest release.

PR #190, opened 2018-05-29
Bump min version of iOS to be 9.0 to follow min version from CrpytoSwift

MikeSilvis:

Update the min version of iOS required to be 9.0.

The min versions conflict with the CrpytoSwift dependency

### DIFF
--- a/PusherSwift.xcodeproj/project.pbxproj
+++ b/PusherSwift.xcodeproj/project.pbxproj
@@ -602,7 +602,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				INFOPLIST_FILE = Tests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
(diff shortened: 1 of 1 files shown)


MikeSilvis:
@jameshfisher any update on this?

edwellbrook:
Hey @jameshfisher @hamchapman. We'd love to see this merged too. Thanks!

apats:
Looks like this will resolve 
https://github.com/pusher/pusher-websocket-swift/issues/194

MikeSilvis:
@apats Yup! I would just use my fork until @jameshfisher or @hamchapman get this merged.

apats:
@MikeSilvis after some investigation I noticed that CryptoSwift has a [PR merged in already](https://github.com/krzyzanowskim/CryptoSwift/pull/632) that sets min required version back to 8.0. Updating dependency to CryptoSwift 0.10.0 seems more reasonable in this case, so iOS 8.0 won't be left behind.

MikeSilvis:
`0.10.1` of Crpyto doesn't include that commit :(

jameshfisher:
@MikeSilvis thanks!

PR #199, opened 2018-11-13
Capture self reference before using self reference in Reachability closures

PorterHoskins:

Implicitly unwrapping `self` is not safe. Checking that `self` is not `nil` is not safe enough to implicitly unwrap `self` later in a function. The only way to make sure `self` is not `nil` is to create a strong reference to `self` before using it.

We are seeing the code crash in our app when the pusher connection is released and the reachability is checked at the same time.

### DIFF
--- a/Sources/PusherConnection.swift
+++ b/Sources/PusherConnection.swift
@@ -42,14 +42,14 @@ public typealias PusherEventJSON = [String: AnyObject]
     open lazy var reachability: Reachability? = {
         let reachability = Reachability.init()
         reachability?.whenReachable = { [weak self] reachability in
-            guard self != nil else {
+            guard let _self = self else {
                 print("Your Pusher instance has probably become deallocated. See https://github.com/pusher/pusher-websocket-swift/issues/109 for more information")
                 return
             }
 
-            self!.delegate?.debugLog?(message: "[PUSHER DEBUG] Network reachable")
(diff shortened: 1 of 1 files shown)


wisron:
@hamchapman Will this be merged anytime soon? My app is crashing occasionally due to wrong captures of self.

Thanks 🙏

TomKemp:
Hi @PorterHoskins. Thanks very much for the PR and sorry for the delay in merging it! It looks good to me. I've just committed a change replacing `_self` with `strongSelf`. I appreciate that we don't have a style guide and really it's just a matter of personal preference, but I saw that some of our dependencies use `strongSelf` (see [here](https://github.com/ashleymills/Reachability.swift/blob/89d8ac704464c3c1f750b7f03797ee0fe7fcebe1/Sources/Reachability.swift#L241)) and some Apple libraries use it too (see [here](https://github.com/search?q=org%3Aapple+strongSelf&type=Code)), so I think that makes sense. I went ahead and committed that change myself seeing as it's been a long time since the PR was made. I'll merge the PR this week unless you can see any problems with that change.

PorterHoskins:
Yeah that is great. Actually now you can just do `guard let self = self else { }`. The swift guys fixed this confusion for us.

TomKemp:
I agree, that would be better. The Swift version is currently set to 4.0 and it doesn't look like that syntax was added until 4.2. Therefore, I will merge this PR as it is for now and the syntax can be updated once the version has been updated.

PR #203, opened 2018-12-19
Allow Newer Dependency Versions

JonathanDowning:

This PR relaxes the pinned versions of PusherSwift's dependencies.

My project is currently unable to update to update CryptoSwift beyond version 0.9.0 and ReachabilitySwift beyond 4.1.0. This is preventing the project from enjoying new features in these libraries such as Swift 4.2 compatibility and bug fixes, to name a few.

I've compiled and run PusherSwift with these changes and I can see no compatibility issues which would prevent this from being a desirable change to make.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-github "ashleymills/Reachability.swift" "v4.1.0"
-github "daltoniam/Starscream" "3.0.5"
+github "ashleymills/Reachability.swift" "v4.3.0"
+github "daltoniam/Starscream" "3.0.6"
 github "icanzilb/TaskQueue" "1.1.1"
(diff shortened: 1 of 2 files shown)


hamchapman:
@jameshfisher this looks good to me 👍

spenrose:
Would love to see this merged. Some of these dependencies are on really old versions of swift.

kevinpalsermtc:
> Would love to see this merged. Some of these dependencies are on really old versions of swift.

I'm having to resort to xySVerma solution: https://stackoverflow.com/questions/40659236/is-it-possible-to-exclude-dependency-in-cocoapods

Please update this so that updating other pod dependencies are not being held back.

imann24:
Hi, +1 to the above comments. Would you mind commenting on the blockers to merging?

PR #204, opened 2019-01-03
`let` properties are implicitly `final`, so the `open` modifier is me…

paulw11:
…aningless.  Change them to `public` in order to avoid a warning from the compiler

Change `let` properties to `public` instead of `open`

Eliminate build warnings in Xcode 10 and later


TomKemp:
Hi @paulw11. Thanks for the PR and sorry for the delay in getting back to you about it. 

The changes in this PR have been merged in a recent PR therefore this PR is redundant and I'll close it.

PR #207, opened 2019-03-01
Remove Push Notifications BETA

lukabratos:

See title.

Please use Pusher Beams, our new Push Notifications product: www.pusher.com/beams
If you're planning to migrate, check our migration guide:
https://pusher.com/docs/push_notifications/migration


iOS Example Obj-C/iOS Example Obj-C/AppDelegate.m
```diff
@@ -19,58 +18,7 @@ @implementation AppDelegate
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     self.pusher = [[Pusher alloc] initWithKey:@"YOUR_APP_KEY"];
```
TomKemp: I don't think the app delegate needs an instance of Pusher. It was only being used for push-notifications.

iOS Example Swift/iOS Example Swift/AppDelegate.swift
```diff
@@ -8,68 +8,15 @@
...
 import UIKit
 import PusherSwift
-import UserNotifications
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate, PusherDelegate {
+class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
     let pusher = Pusher(key: "YOUR_APP_KEY")
```
TomKemp: Same here. I don't think the app delegate needs an instance of `Pusher`. It was only being used for push-notifications.

macOS Example Swift/macOS Example Swift/AppDelegate.swift
```diff
@@ -10,55 +10,15 @@ import Cocoa
 import PusherSwift
 
 @NSApplicationMain
-class AppDelegate: NSObject, NSApplicationDelegate, PusherDelegate {
+class AppDelegate: NSObject, NSApplicationDelegate {
 
     let pusher = Pusher(key: "YOUR_APP_KEY")
```
TomKemp: Same again. I don't think the app delegate needs an instance of Pusher. It was only being used for push-notifications.

iOS Example Obj-C/iOS Example Obj-C/AppDelegate.m
lukabratos: You're correct. We just had example code related to push notification here.
I have intentionally left at least instance of Pusher client in here if we would potentially add more examples - maybe related to some other functionality that this SDK provides but we can do that later. I'll remove this code.

Sources/PusherSwift.swift
```diff
@@ -11,16 +11,11 @@ let CLIENT_NAME = "pusher-websocket-swift"
...
         willSet {
             self.connection.delegate = newValue
-#if os(iOS) || os(OSX)
-            self.nativePusher.delegate = newValue
-#endif
         }
     }
     private let key: String
 
 #if os(iOS) || os(OSX)
```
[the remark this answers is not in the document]
lukabratos: Nice catch, thanks 👍

TomKemp:
Just leaving a note so I remember to check further before merging this: I think the `TaskQueue` dependency is only used for the old native push notifications, so that dependency can be removed.

PR #209, opened 2019-03-13
Fix unreliable tests

TomKemp:
This PR fixes a number of unreliable tests that sometimes fail (but not consistently). Each of the tests depend on the order of a dictionary. Dictionaries are unordered and their orders are unpredictable,  leading to inconsistent test results.

In the case of `testUnsubscribingFromAllChannelsRemovesTheChannels`, the test depended on `test-channel2` being unsubscribed from last. However, the order in which the channels are unsubscribed from depends on the order of the channels dictionary, which is unpredictable. The test was fixed by checking that unsubscribe messages were sent for both channels, ignoring the order.

In the case of `testFindingTheClientsMemberObject`, `testFindingAPresenceChannelAsAPusherPresenceChannel` and `testMembersObjectStoresUserIdAndUserInfoIfAUserDataFetcherIsProvidedThatReturnsBoth`, the auth signature depends on the order of the keys in the channel_data JSON. The order of the keys in the channel_data JSON depends on the order of the dictionary from which it is generated, which is unpredictable. These tests are fixed by adding extra auth signatures for the other possible orderings. It may be beneficial to modify the mocks to verify the auth signatures, rather than hardcode them, but that is a more significant change and could be considered for future work.

### DIFF
--- a/Tests/PusherTopLevelAPITests.swift
+++ b/Tests/PusherTopLevelAPITests.swift
@@ -241,13 +241,14 @@ class PusherTopLevelApiTests: XCTestCase {
             let unsubscribedFromChannel = socket.stubber.calls.contains { call in
                 guard
                     call.name == "writeString",
-                    let firstArg = call.args?.first
+                    let arguments = call.args,
+                    let firstArg = arguments.first,
+                    let stringFirstArg = firstArg as? String,
+                    let parsedCallArgs = convertStringToDictionary(stringFirstArg)
                 else {
                     return false
                 }
-
-                let parsedCallArgs = convertStringToDictionary(firstArg as! String)
-                return NSDictionary(dictionary: parsedCallArgs!).isEqual(to: NSDictionary(dictionary: expectedCallArguments) as [NSObject: AnyObject])
+                return NSDictionary(dictionary: parsedCallArgs).isEqual(to: NSDictionary(dictionary: expectedCallArguments) as [NSObject: AnyObject])
             }
             XCTAssertTrue(unsubscribedFromChannel, "should have unsubscribed from \(channel)")
         }
--- a/Tests/PusherTopLevelAPITests.swift
+++ b/Tests/PusherTopLevelAPITests.swift
@@ -226,19 +226,32 @@ class PusherTopLevelApiTests: XCTestCase {
 
     func testUnsubscribingFromAllChannelsRemovesTheChannels() {
         pusher.connect()
-        let _ = pusher.subscribe("test-channel")
-        let _ = pusher.subscribe("test-channel2")
-        XCTAssertEqual(pusher.connection.channels.channels.count, 2, "should have 2 channels")
+        let channels = ["test-channel", "test-channel2"]
+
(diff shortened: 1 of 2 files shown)


Tests/PusherTopLevelAPITests.swift
```diff
@@ -226,19 +226,32 @@ class PusherTopLevelApiTests: XCTestCase {
...
+            let expectedCallArguments = ["data": ["channel": channel], "event": "pusher:unsubscribe"] as [String: Any]
+            let unsubscribedFromChannel = socket.stubber.calls.contains { call in
+                guard
+                    call.name == "writeString",
+                    let firstArg = call.args?.first
+                else {
+                    return false
+                }
+
+                let parsedCallArgs = convertStringToDictionary(firstArg as! String)
```
lukabratos: Are we sure that `firstArg` will always be of a type `String`?

Tests/PusherTopLevelAPITests.swift
```diff
@@ -226,19 +226,32 @@ class PusherTopLevelApiTests: XCTestCase {
...
 
-        let parsedSubscribeArgs = convertStringToDictionary(socket.stubber.calls.last?.args!.first as! String)
-        let expectedDict = ["data": ["channel": "test-channel2"], "event": "pusher:unsubscribe"] as [String: Any]
-        let parsedEqualsExpected = NSDictionary(dictionary: parsedSubscribeArgs!).isEqual(to: NSDictionary(dictionary: expectedDict) as [NSObject: AnyObject])
+        for channel in channels {
+            let expectedCallArguments = ["data": ["channel": channel], "event": "pusher:unsubscribe"] as [String: Any]
+            let unsubscribedFromChannel = socket.stubber.calls.contains { call in
+                guard
+                    call.name == "writeString",
+                    let firstArg = call.args?.first
```
lukabratos: I would expand this and handle the optional. Something like this:

```swift
call.name == "writeString",
let arguments = call.args,
let firstArg = arguments.first
```

Tests/PusherTopLevelAPITests.swift
```diff
@@ -226,19 +226,32 @@ class PusherTopLevelApiTests: XCTestCase {
...
+                else {
+                    return false
+                }
+
+                let parsedCallArgs = convertStringToDictionary(firstArg as! String)
+                return NSDictionary(dictionary: parsedCallArgs!).isEqual(to: NSDictionary(dictionary: expectedCallArguments) as [NSObject: AnyObject])
```
lukabratos: I would avoid force unwrapping `parsedCallArgs`.

Tests/PusherTopLevelAPITests.swift
TomKemp: Within this test it will always be a string, but I've removed the forced cast in the latests commit.

TomKemp:
Thanks for the review. What do you think of the changes in the latest commit?

TomKemp:
Luka and I reviewed this together in person and decided it's ready to merge.

PR #210, opened 2019-03-25
Fix error caused by multiple 'fulfill' calls on expectations

TomKemp:
Reachability was calling the 'whenReachable' callback straight after connecting, which led to reconnection attempts. Those reconnections caused certain tests to throw exceptions because 'fulfill' was called multiple times on expectations. E.g.:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'API violation - multiple calls made to -[XCTestExpectation fulfill] for the channel should be subscribed to successfully.'
```

This PR disables autoReconnect for tests where 'connect' is called, so that Reachability isn't activated. These tests aren't specifically testing the reconnection behaviour or the behaviour of Reachability. Having those features enabled makes the tests less reliable and makes the tests dependent on network connectivity. The reconnection logic should be tested in separate tests.

### DIFF
--- a/Tests/PresenceChannelTests.swift
+++ b/Tests/PresenceChannelTests.swift
@@ -11,7 +11,8 @@ class PusherPresenceChannelTests: XCTestCase {
         super.setUp()
 
         options = PusherClientOptions(
-            authMethod: .inline(secret: "secret")
+            authMethod: .inline(secret: "secret"),
+            autoReconnect: false
         )
         pusher = Pusher(key: "key", options: options)
         socket = MockWebSocket()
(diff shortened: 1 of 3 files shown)


Tests/PresenceChannelTests.swift
```diff
@@ -11,7 +11,8 @@ class PusherPresenceChannelTests: XCTestCase {
         super.setUp()
 
         options = PusherClientOptions(
```
WillSewell: Any reason why the others have a `let` whereas this does not?

TomKemp: This is setting the `options` property on the class which is used by one of the tests. The other examples in this file are creating their own instance of `PusherClientOptions` with custom parameters for their tests. I'm not sure why those tests needed different secrets. It might be something to do with requiring different behaviour in the mocks, but I would need to take a closer look.

WillSewell: Ah right yeah. I didn't spend enough looking at the context of this line of code.

PR #212, opened 2019-03-29
Defining versions in Cartfile for dependencies 

cowgp:
Defining versions in Cartfile for dependencies so dependencies can update to latest xcode/swift and publish breaking versions without breaking current release of this when still run on something other than the latest xcode/swift development environment

Reachability, Starscream, and CryptoSwift have all published releases compatible with Swift 5 and XCode 10.2.   Because the current Cartfile does not define any version requirements for the depencies, the "latest" release is always taken.    Running `carthage update` or `carthage bootstrap` as our CI build tool does is now failing as we are not yet running XCode 10.2 in CI (or locally).   

Going forward, as releases of this library come out supporting different swift/xcode versions, the versions of the dependencies should remain defined (and updated) in the Cartfile so that they build with the correct known working versions.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
-github "krzyzanowskim/CryptoSwift"
-github "ashleymills/Reachability.swift"
-github "icanzilb/TaskQueue"
-github "daltoniam/Starscream"
+github "krzyzanowskim/CryptoSwift" ~> 0.15
+github "ashleymills/Reachability.swift" == 4.3.0
+github "icanzilb/TaskQueue" ~> 1.0.3
+github "daltoniam/Starscream" == 3.0.6
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
-github "krzyzanowskim/CryptoSwift" ~> 0.15
+github "krzyzanowskim/CryptoSwift" == 0.15.0
 github "ashleymills/Reachability.swift" == 4.3.0
-github "icanzilb/TaskQueue" ~> 1.0.3
+github "icanzilb/TaskQueue" == 1.1.1
 github "daltoniam/Starscream" == 3.0.6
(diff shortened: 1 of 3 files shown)


TomKemp:
Thanks for the PR. That seems sensible. What was the reason for lowering the version of TaskQueue?

Also, the dependencies in the Podspec should match the Cartfile, so I'll make some changes to the Podspec to match and push them to this branch.

cowgp:
I made the podspec and carthage consistent and switched all dependencies to == rather than ~> as CryptoSwift was somehow evaluating to 1.0.0 when it was set to ~> 0.15 which shouldn't be possible, but it was.  I did put TaskQueue back to 1.1.1 - when I went through and put the values in for the dependencies I was looking at the releases on each repo, and TaskQueue hasn't been publishing releases even though they are tagging each one - so I put an old version in as a result.  But it's now at listed as 1.1.1

TomKemp:
Sorry, I didn't get round to pushing my changes. It looks like we ended up with roughly the same Podspec though. The only difference is that I don't think it's necessary to pin CryptoSwift and TaskQueue to precise versions. It's probably desirable to allow flexibility.

I've tested my version using both Carthage and CocoaPods and I get the correct versions installed. I noticed that `~>`to means something different in Carthage vs Cocoapods, which could be why you were having problems, but I'm not sure how that would end up returning `1.0.0`.

I've pushed my proposed changes. What do you think?

TomKemp:
I appreciate that the versions don't match exactly between the Cartfile and Podspec but my reasoning is that:

- Starscream: updated to Swift 5 on a minor version. In Carthage `~>` will update minor versions so we have to pin to an exact version. In CocoaPods however, we can use `~>` to only update patch versions. It makes sense for us to allow flexibility where we can, whilst still holding the version back before that Swift 5 release.

- CryptoSwift: the Swift 5 release was 1.0.0, but the behaviour of `~>` in Carthage is different in versions < 1.0.0 and won't update minor versions, so we can't express the same restrictions in Carthage and CocoaPods.

- Reachability: updated to Swift 5 on a patch version therefore we have to pin to an exact version in both files.

- TaskQueue: no recent changes.

TomKemp:
@lukabratos and I reviewed this together in person and decided it's ready to merge.

PR #214, opened 2019-04-09
Update to Swift 4.2

TomKemp:
Upgrade the Swift version to 4.2. As part of the upgrade, this PR fixes compiler warnings that result in complier errors when using the library with Swift 5 and sets the Swift version in the Podspec so CocoaPods compiles it as Swift 4.2.


TomKemp:
@lukabratos and I reviewed this in person and decided that it's ready to merge.

PR #224, opened 2019-07-25
Update Package.swift

JonathanDowning:

The current Package.swift is malformed and does not work with Xcode 11's SPM integration.
This PR fixes this.

I've ensured that these changes and the pinned dependency versions are Swift 4.2 compatible.

The tests needed to be modified slightly to compile also.

Without this change, PusherSwift is not compatible with Swift Package Manager


JonathanDowning:
How do you feel about this addition @TomKemp? :)

lordcodes:
@TomKemp Please can we get this PR merged?

Now that Xcode 11 is out of beta, many developers will be moving dependencies across to SPM. Without this PR merged, that is currently not possible with PusherSwift.

JonathanDowning:
+1 :P

FradSer:
😅Waitting

TomKemp:
Hi @JonathanDowning. Thanks a lot for the PR and sorry for the delay in getting back to you about it! It looks good but I've pushed some changes. 

Firstly, I'd like to try to keep the dependencies in the Package.swift/Podspec/Cartfile as close as possible to make maintenance easier. Therefore I've updated the Package.swift to match the Podspec more closely. I had to change the Starscream version because `3.0.5` has an incompatible Package.swift.

Secondly, I removed Swift 5 as a language version. Again, I did that to match the Podspec. I would like to update the library so it can be compiled as Swift 5, but I think it would be best to do that in a separate PR that also updates the library to remove deprecated syntax.

What do you think of those changes?

JonathanDowning:
That's all fine by me @TomKemp, thanks for your input on this!

TomKemp:
Great, thank you 🙂

PR #237, opened 2019-10-24
Add Swift 5 Support

JonathanDowning:

This PR drops support for Swift 4.2, updates dependencies to latest versions, and changes syntax to be compatible with Swift 5 and the newer dependencies.

PusherSwift currently uses older versions of dependencies to maintain support for Swift 4.2, since Swift 5.1 is now the latest, it's perhaps high time to drop 4.2.


JonathanDowning:
Any thoughts @TomKemp ? :)

JonathanDowning:
Happy new year bump

TomKemp:
Hi @JonathanDowning. Thanks very much for the PR!

I've started going through it. I'll post a further update tomorrow.

TomKemp:
The PR looks good. There are two things that I will address next week:

- The semantics of the "~>" are slightly different in Cocoapods vs Carthage. The PR currently allows Cocoapods to update patch versions only, whereas Carthage can update minor versions. I'll change the versions so they're consistent.

- Version 1.2.0 of CryptoSwift actually uses Swift 5.1, so users of the library would need to use an Xcode version with the Swift 5.1 compiler available rather than just Swift 5.0. And the library's dependencies seem to update the Swift version in minor/patch versions. Therefore I think either the dependencies will need to be pinned to specific versions, or the README will need to encourage users to use the latest version of Xcode, rather than specifying a minimum Xcode version as it does now. I think encouraging users to use the latest version of Xcode is the better option, but I will discuss it with my colleagues next week and come up with a policy.

JonathanDowning:
Hey @TomKemp thank you so much for looking at this, I really appreciate it!
I would be strongly in favor of keeping up with the latest swift versions and indicating to developers the last supported swift version [like CryptoSwift does](https://github.com/krzyzanowskim/CryptoSwift#swift-versions-support). As you mentioned. 
Have a great weekend!

JonathanDowning:
Closing due to #261, thanks @TomKemp !

PR #245, opened 2020-02-17
Replace CryptoSwift with CommonCrypto HMAC functions

TomKemp:
Currently we use CryptoSwift to generate a HMAC when users are using the ".inline" auth method for testing. CryptoSwift is a large library and takes a long time to compile. Furthermore, it is desirable to have fewer dependencies because it is difficult to manage multiple dependencies across the multiple package managers that we support.

This PR removes CryptoSwift and instead uses the `CommonCrypto` APIs.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,8 +1,2 @@
-# CryptoSwift versions 0.9 - 0.10 inclusive have iOS deployment
-# target "9.0" set in the Xcode project (but not in the Podspec).
-# PusherSwift has target "8.0" therefore the version in this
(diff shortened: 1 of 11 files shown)


OliverPearmain-Pusher:
I think the twp demo apps might also need updating  because they explcitly copy the `CryptoSwift.framework` in their "Copy Carthage Frameworks" build phase.

TomKemp:
Thanks @OliverPearmain-Pusher. I had made the changes but hadn't committed them. That should be better now

PR #247, opened 2020-03-24
E2E Decrypt

OliverPearmain-Pusher:

- Added Carthage and `Swift-Sodium` dependency to main project target
-  Implemented one half of E2E encryption. This part decrypts the payload from an event but at present its necessary to hardcode your decryption key which in future will be provided in the auth response from the server.
- Switched some of the computed properties on `PusherEvent` to be `lazy` loaded so they aren't repeatedly computed and performance is improved.
- Added additional `PusherEventTests` to cover decryption.

- So that the SDK ultimately supports E2E encryption

#### Whats Outstanding?
- The parts of the code where things integrate with the auth call.  I have added a couple of `TODO` s that hopefully cover this.

#### How to See It Working in the Demo App
(This assumes you have a server setup and running somewhere already)

1.  Update `iOS Example Swift/ViewController.swift` with the following...
```
        let pusherHost = PusherHost.cluster("eu")
        let pusherClientOptions = PusherClientOptions(authMethod: .inline(secret: "<YOUR_SECRET>"), host: pusherHost)
        pusher = Pusher(key: "<YOUR_KEY>", options: pusherClientOptions)
```
```
        let myChannel = pusher.subscribe("private-encrypted-<YOUR-CHANNEL>")
```
2. Update `Sources/PusherConnection.swift` hardcoding your decryptionKey here...
```
    internal var keyProvider: PusherKeyProviding {
        let decryptionKey = "<ENTER_YOUR_KEY>"
        return PusherKeyProvider(decryptionKey: decryptionKey)
    }
```
3.  Run the `iOS Example` scheme onto a simulator.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -6,3 +6,4 @@
 github "krzyzanowskim/CryptoSwift" ~> 0.15.0
 github "ashleymills/Reachability.swift" == 4.3.0
 github "daltoniam/Starscream" == 3.0.6
+github "jedisct1/swift-sodium" == 0.8.0
--- a/PusherSwift.podspec
+++ b/PusherSwift.podspec
@@ -15,6 +15,7 @@ Pod::Spec.new do |s|
   s.dependency 'CryptoSwift', '~> 0.9'
   s.dependency 'ReachabilitySwift', '4.3.0'
   s.dependency 'Starscream', '~> 3.0.5'
+  s.dependency 'Sodium', '~> 0.8.0'
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.10'
--- a/PusherSwift.xcodeproj/project.pbxproj
+++ b/PusherSwift.xcodeproj/project.pbxproj
@@ -32,6 +32,7 @@
 		33BB997C1D21230100B25C2A /* PusherTopLevelAPITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33BB99661D21226C00B25C2A /* PusherTopLevelAPITests.swift */; };
 		33C1FD6F1D81BFC300921AD7 /* ObjectiveC.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33C1FD6E1D81BFC300921AD7 /* ObjectiveC.swift */; };
 		33C40CB91C1DFC9C008A54E3 /* PusherSwift.h in Headers */ = {isa = PBXBuildFile; fileRef = 33831CD61A9CFFF200B124F1 /* PusherSwift.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		730EECB724290BE700F0EB73 /* Sodium.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 730EECB624290BE700F0EB73 /* Sodium.framework */; };
 		E2498293231E612700CFBBD6 /* PusherError.swift in Sources */ = {isa = PBXBuildFile; fileRef = E2498292231E612700CFBBD6 /* PusherError.swift */; };
 		E24D0AA522D798C8009DE31B /* PusherEvent.swift in Sources */ = {isa = PBXBuildFile; fileRef = E24D0AA422D798C8009DE31B /* PusherEvent.swift */; };
 		E29A4CBC2301BA31000BC499 /* PusherEventTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E29A4CBB2301BA31000BC499 /* PusherEventTests.swift */; };
@@ -78,6 +79,7 @@
 		33BB99661D21226C00B25C2A /* PusherTopLevelAPITests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = PusherTopLevelAPITests.swift; path = ../Tests/PusherTopLevelAPITests.swift; sourceTree = "<group>"; };
 		33BB99671D21226C00B25C2A /* Info.plist */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.xml; name = Info.plist; path = ../Tests/Info.plist; sourceTree = "<group>"; };
 		33C1FD6E1D81BFC300921AD7 /* ObjectiveC.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ObjectiveC.swift; sourceTree = "<group>"; };
+		730EECB624290BE700F0EB73 /* Sodium.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = Sodium.framework; path = Carthage/Build/iOS/Sodium.framework; sourceTree = "<group>"; };
 		E2498292231E612700CFBBD6 /* PusherError.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PusherError.swift; sourceTree = "<group>"; };
 		E24D0AA422D798C8009DE31B /* PusherEvent.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PusherEvent.swift; sourceTree = "<group>"; };
 		E29A4CBB2301BA31000BC499 /* PusherEventTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PusherEventTests.swift; sourceTree = "<group>"; };
@@ -96,6 +98,7 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				730EECB724290BE700F0EB73 /* Sodium.framework in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -177,6 +180,7 @@
 		42D29A655054030A9B238900 /* Frameworks */ = {
 			isa = PBXGroup;
 			children = (
+				730EECB624290BE700F0EB73 /* Sodium.framework */,
 			);
 			name = Frameworks;
 			sourceTree = "<group>";
@@ -211,6 +215,7 @@
 				33831C861A9CF61600B124F1 /* Headers */,
 				33831C871A9CF61600B124F1 /* Resources */,
 				50355EA8A07B4B8038A4DDF1 /* Frameworks */,
+				730EECB52429067100F0EB73 /* Copy Carthage Dependencies */,
 			);
 			buildRules = (
 			);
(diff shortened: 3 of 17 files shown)


Cartfile
```diff
@@ -6,3 +6,4 @@
 github "krzyzanowskim/CryptoSwift" ~> 0.15.0
 github "ashleymills/Reachability.swift" == 4.3.0
 github "daltoniam/Starscream" == 3.0.6
+github "jedisct1/swift-sodium" == 0.8.0
```
grzesiekko: Carthage integration is fine, but there are 4 potential issues that we might need to solve:

1. Developers can integrate with the SDK using CocoaPods, and therefore we should declare the same dependency in `PusherSwift.podspec` file.
2. The same applies to SPM support and `Package.swift` file.
3. `Sodium` supports iOS, but the framework is not marked as safe to use in iOS extensions. This means that we either need to mark our SDK as not safe to use in iOS extensions or find an alternative solution/fix.
4. `Sodium` does not support `tvOS` and our SDK does. We can either mark the feature as not available on `tvOS` or find an alternative library/solution.

OliverPearmain-Pusher:
FYI I've just pushed an `e2e` branch (based off of `master`) and changed the destination of this brach to `e2e`.

Cartfile
OliverPearmain-Pusher: Thanks Greg, all valid points and things I'd overlooked. 

I have pushed a change to address 1.

WRT 2. 3. & 4. we have since discussed this and decided to defer this work for another PR (now that this one is targetting the `e2e` intermediate branch and not `master`).

OliverPearmain-Pusher:
This is ready for re-review please.  Changes Ive recently made:

- Updated code so that decryption is not attempted for event names prefixed "pusher:" or "pusher_internal:"
- sorted the Cocoapods integration
- Reverted the addition of "Carthage Copy Frameworks" build phases from the main target (this assumes that the client is using Carthage which is not always the case).

PusherSwift.xcodeproj/project.pbxproj
```diff
@@ -78,6 +83,11 @@
...
 		33C1FD6E1D81BFC300921AD7 /* ObjectiveC.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ObjectiveC.swift; sourceTree = "<group>"; };
+		730EECB624290BE700F0EB73 /* Sodium.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = Sodium.framework; path = Carthage/Build/iOS/Sodium.framework; sourceTree = "<group>"; };
+		736E53F3242A35D90052CC1B /* String+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+Extensions.swift"; sourceTree = "<group>"; };
+		736E53F6242A45AC0052CC1B /* XCTest+Assertions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCTest+Assertions.swift"; sourceTree = "<group>"; };
+		736E53F8242A4DE60052CC1B /* PusherKeyProvider.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PusherKeyProvider.swift; sourceTree = "<group>"; };
+		736E53FC242A55230052CC1B /* PusherKeyProvider.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PusherKeyProvider.swift; sourceTree = "<group>"; };
```
mdpye: This line appears duplicated (and the same in the block above)

OliverPearmain-Pusher: Thats because there are 2 `PusherKeyProvider.swift` files...
- `Sources/PusherKeyProvider.swift`
- `Tests/Doubles/PusherKeyProvider.swift`

PusherSwift.xcodeproj/project.pbxproj
```diff
@@ -478,6 +506,10 @@
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(PROJECT_DIR)/Carthage/Build/iOS",
```
grzesiekko: Should this be hardcoded to `iOS` or should this platform specific?

OliverPearmain-Pusher: I have removed this addition altogether.  Its not required because the existing `Carthage.xcconfig` covers it.

Sources/PusherEvent.swift
```diff
@@ -1,26 +1,96 @@
...
+            return nil
+        }
+        return [UInt8](decoded_data_cipherText)
+    }()
 
-    @nonobjc internal init?(jsonObject: [String: Any]) {
+    @nonobjc private let sodium = Sodium()
+    @nonobjc private let keyProvider: PusherKeyProviding?
+    
+    @nonobjc internal init?(jsonObject: [String: Any], keyProvider: PusherKeyProviding? = nil) {
```
TomKemp: I'm not sure about the decryption logic living inside `PusherEvent`.  I think of `PusherEvent` as being a simple data structure that stores the event data with a few simple helper functions inside. It seems a bit wrong to pass that a key provider and for each event to have a copy of the secret key inside it. I was considering having some kind of decrypter function that you pass a `PusherEvent` with encrypted `data` and it returns a new `PusherEvent` with `data` decrypted. I guess that isn't the most efficient and I see the benefit of the current implementation lazily decrypting messages.

I don't feel too strongly about it and I think this implementation is pretty neat so I don't mind if it stays the way it is, but I just wanted to make the point.

Tests/PusherEventTests.swift
```diff
@@ -133,4 +132,258 @@ class PusherEventTests: XCTestCase {
...
+        }
+        """.removing(.whitespacesAndNewlines)
+        
+        let jsonDict = """
+        {
+            "event": "test-event",
+            "channel": "my-channel",
+            "data": \(dataPayload.escaped)
+        }
+        """.toJsonDict()
```
TomKemp: I just wanted to say that I really like this structure 🙂

Cartfile
```diff
@@ -1,2 +1,3 @@
...
```
TomKemp: I think version 0.8.0 uses Swift 5, so we'll probably have to change the README to list Xcode 10.2 rather than 10. Also the MacOS deployment target is 10.11 whereas the current target for PusherSwift is 10.10 so we'll have to change that.

PusherSwift.podspec
```diff
@@ -14,6 +14,7 @@ Pod::Spec.new do |s|
 
   s.dependency 'ReachabilitySwift', '4.3.0'
   s.dependency 'Starscream', '~> 3.0.5'
+  s.dependency 'Sodium', '~> 0.8.0'
```
TomKemp: Why does this allow patch version increases but the Cartfile is pinned to a specific version?

Sources/PusherEvent.swift
OliverPearmain-Pusher: I think you raise a valid point.  However I'm not sure I agree that the existing `PusherEvent` is a simple data structure because it performs JSON pasring internally.  

I think in an ideal world a `PusherEvent` would truely be a data/value type and there would be a seperate JSON parser that would perform the decryption of data when necessary (probably using an injected decrypter).  This would however be a significant refactor so I don't think its something we can feasibly consider now.

Tests/PusherEventTests.swift
OliverPearmain-Pusher: Thanks, the aim is to keep the tests as readable as possible.  We have tried to use a very the same appraoch in the chatkit-swift tests (in the version-2.0 branch).

Sources/PusherEvent.swift
TomKemp: Yeah, the JSON parsing probably shouldn't be in there. That's fair enough 👍

Cartfile
OliverPearmain-Pusher: Good point.  TBH the project and CI should be setup in a way such that compilation fails if changes violate any of the integration requirements.  Really and truely we need a suite of test integration projects that all get compiled in order for CI to go green.  

There would be quite a few projects however, for example...

-  Mac app, using Carthage and lowest supported version of Xcode & Swift
-  Mac app, using Cocoapods and lowest supported version of Xcode & Swift
-  Mac app, using SPM and lowest supported version of Xcode & Swift
-  Mac app, using Carthage and latest version of Xcode & Swift
-  Mac app, using Cocoapods and latest version of Xcode & Swift
-  Mac app, using SPM using and supported version of Xcode & Swift

-  iOS app with extensions, using Carthage and lowest supported version of Xcode & Swift
-  iOS app with extensions, using Cocoapods and lowest supported version of Xcode & Swift
-  iOS app with extensions, using SPM and lowest supported version of Xcode & Swift
-  iOS app with extensions, using Carthage and latest version of Xcode & Swift
-  iOS app with extensions, using Cocoapods and latest version of Xcode & Swift
-  iOS app with extensions, using SPM using and supported version of Xcode & Swift

-  tvOS app, using Carthage and lowest supported version of Xcode & Swift
-  tvOS app, using Cocoapods and lowest supported version of Xcode & Swift
-  tvOS app, using SPM and lowest supported version of Xcode & Swift
-  tvOS app, using Carthage and latest version of Xcode & Swift
-  tvOS app, using Cocoapods and latest version of Xcode & Swift
-  tvOS app, using SPM using and supported version of Xcode & Swift

PusherSwift.podspec
OliverPearmain-Pusher: Its a good question.  The answer not so much... because I copied and pasted.   

The same question applies to the existing dependencies.

TomKemp: Fair enough. For Starscream, it's because version `3.1.0` adds Swift 5 support, which we were avoiding because it's not available in Xcode 10. It's possible to express patch version compatibility in Cocoapods, but not in Carthage. So I pinned Carthage to a specific version and allowed patch version updates in Cocoapods to avoid dependency conflicts for customers as much as possible. We really need a better policy for Swift version compatibility though because it's impossible to manage all 3 dependency managers in a consistent way when Swift versions of dependencies can change in minor/patch versions.

OliverPearmain-Pusher: To further complicate things, I don't think Cocoapod's and Carthage's implementations of the `~>` operator are equivalent. 

[Cocoapod's docs](https://guides.cocoapods.org/using/the-podfile.html) are pretty clear...
```
    '~> 0.1.2' Version 0.1.2 and the versions up to 0.2, not including 0.2 and higher
    '~> 0.1' Version 0.1 and the versions up to 1.0, not including 1.0 and higher
    '~> 0' Version 0 and higher, this is basically the same as not having it.
```

[Carthage's docs](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md) are no so great...
```
~> 1.0 for “compatible with version 1.0”

Compatibility is determined according to Semantic Versioning. This means that any version greater than or equal to 1.5.1, but less than 2.0, will be considered “compatible” with 1.5.1.

According to SemVer, any 0.x.y release may completely break the exported API, so it's not safe to consider them compatible with one another. Only patch versions are compatible under 0.x, meaning 0.1.1 is compatible with 0.1.2, but not 0.2. This isn't according to the SemVer spec but keeps ~> useful for 0.x.y versions.
```

OliverPearmain-Pusher: Having just done some Carthage tests it seems to ignore any patch number (the third value)  specified.

So if you use `~> 0.1.2` its the equivalent of using `~> 0.1`.

TomKemp: Yeah, that's my understanding of how it works.

OliverPearmain-Pusher: > Swift versions of dependencies can change in minor/patch versions.

If thats the case then should we just pin all the dependencies?

grzesiekko: Developers using Channels SDK might be also using the same dependencies directly in their projects. Giving them some flexibility (`~>`) helps in avoiding a situation of clashing dependency versions.

grzesiekko: > Having just done some Carthage tests it seems to ignore any patch number (the third value) specified.
> 
> So if you use `~> 0.1.2` its the equivalent of using `~> 0.1`.

I can confirm that this is a know issue/feature of Carthage that it ignores patch version when referenced using `~>` operator.

TomKemp: I'm not sure what the best solution is to this. It's been a problem for a while. I was leaning towards encouraging developers to use this SDK with the latest version of Xcode and removing the Xcode version from the README. Then we would only have to worry about making sure the SDK and dependencies support the latest version of Xcode. And we could unpin the dependency versions. I think I've seen more complaints about outdated dependencies than lack of support for older versions of Xcode.

But before resorting to that, I was attempting to reduce the number of dependencies, hence why that PR to remove CryptoSwift existed. But I don't think it will be possible to get rid of them all, so the issue still exists.

OliverPearmain-Pusher: @TomKemp in the interests of getting this merged to the integration branch, are you happy for me to take this is as task to resolve in my current work and my next PR?

(I am addressing a number of related issues, depedency support for tvOS, SPM, etc)

TomKemp: Yeah sure 👍

PR #255, opened 2020-04-16
Additional unit tests + EventQueue simplification

TomKemp:
- Adds tests for:
     - Key provider
    - Event queue
    - Event factory

- Simplifies the EventQueue code

### DIFF
--- a/Tests/PusherEventFactoryTests.swift
+++ b/Tests/PusherEventFactoryTests.swift
@@ -206,4 +206,17 @@ class PusherEventFactoryTests: XCTestCase {
         XCTAssertTrue(event.property(withKey: "my_null") is NSNull)
     }
 
+    func testInvalidMessageThrowsException() {
+        let jsonDict = """
+        {
+            "channel": "my-channel",
+            "data": "{\\"test\\":\\"test string\\",\\"and\\":\\"another\\"}"
+        }
+        """.toJsonDict();
+
+        XCTAssertThrowsError(try eventFactory.makeEvent(fromJSON: jsonDict, withDecryptionKey: nil)) { (error) in
+            XCTAssertEqual(error as? PusherEventError, PusherEventError.invalidFormat)
+        }
+    }
+
 }
--- a/Tests/PusherEventFactoryTests.swift
+++ b/Tests/PusherEventFactoryTests.swift
@@ -6,7 +6,7 @@ import XCTest
     @testable import PusherSwift
 #endif
 
-class PusherConcreteEventFactoryTests: XCTestCase {
+class PusherEventFactoryTests: XCTestCase {
 
     var eventFactory: PusherConcreteEventFactory!
 
--- a/Tests/PusherEventQueueTests.swift
+++ b/Tests/PusherEventQueueTests.swift
@@ -1,0 +1,97 @@
+import XCTest
+
+#if WITH_ENCRYPTION
+    @testable import PusherSwiftWithEncryption
+#else
+    @testable import PusherSwift
+#endif
+
+class InlineMockEventQueueDelegate: PusherEventQueueDelegate {
+    var didReceiveEvent: ((PusherEventQueue, PusherEvent, String?) -> Void)?
+    var didFailToDecryptEvent: ((PusherEventQueue, PusherEventPayload, String?) -> Void)?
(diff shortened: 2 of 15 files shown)


Tests/PusherEventFactoryTests.swift
```diff
@@ -206,4 +206,17 @@ class PusherConcreteEventFactoryTests: XCTestCase {
         XCTAssertTrue(event.property(withKey: "my_null") is NSNull)
     }
 
+    func testInvalidMessageThrowsException() {
+        let jsonDict = """
```
daniellevass, line 210: is this message invalid because it uses two escape chars per quote e.g. `\\"`?

daniellevass: or because it doesn't have an event property?

Tests/PusherEventQueueTests.swift
```diff
@@ -0,0 +1,127 @@
...
+            XCTAssertTrue(equal)
+            XCTAssertEqual("my-channel", channelName)
+            ex.fulfill()
+        }
+
+        eventQueue.enqueue(json: jsonDict)
+        wait(for: [ex], timeout: 0.5)
+    }
+
+    func testEventWithNoChannelShouldCallDidReceiveEvent() {
```
daniellevass, line 76: does this mean if a message doesn't have a channel, the client will receive it? how is that not an invalid event?

Tests/PusherEventQueueTests.swift
```diff
@@ -0,0 +1,127 @@
...
+            XCTAssertNil(channelName)
+            ex.fulfill()
+        }
+
+        eventQueue.enqueue(json: jsonDict)
+        wait(for: [ex], timeout: 0.5)
+    }
+
+    func testInvalidEventShouldCallDidReceiveInvalidEvent() {
+        let dataPayload = """
```
daniellevass, line 104: is this one invalid because it doesn't have an event & data payload?

Tests/PusherSwift-Only/PusherEventQueue+DecryptionTests.swift
```diff
@@ -0,0 +1,79 @@
...
+        let ex = expectation(description: "should call didReceiveEvent")
+        eventQueueDelegate.didReceiveEvent = { (eventQueue, event, channelName) in
+            let equal = NSDictionary(dictionary: jsonDict).isEqual(to: event.raw)
+            XCTAssertTrue(equal)
+            XCTAssertEqual("private-encrypted-channel", channelName)
+            ex.fulfill()
+        }
+
+        eventQueue.enqueue(json: jsonDict)
+        wait(for: [ex], timeout: 0.5)
```
daniellevass, line 51: is your ex in a square box because you can provide a list of expectations here? Is this better than the `waitForExpectations`?

Tests/PusherEventFactoryTests.swift
TomKemp: It's invalid because it doesn't have an event. The extra escaping seems to be necessary in the `"""` block.

Tests/PusherEventQueueTests.swift
TomKemp: Some events aren't channel specific, for example: `pusher:connection_established`

Tests/PusherEventQueueTests.swift
TomKemp: That one is also invalid because it doesn't have an `event`.

Tests/PusherSwift-Only/PusherEventQueue+DecryptionTests.swift
TomKemp: Apparently `waitForExpectations` waits on expectations created with the XCTest `expectation` helper method, so if you want to wait expectations created directly with `XCTestExpectation`'s constructor then you have to use `wait(for: [ex], timeout: 0.5)` and pass the expectation. I've just pushed a commit changing the `wait`s to `waitForExpectations` so it's more consistent.

daniellevass: Ah. That's interesting and also slightly confusing! But good to know :) Thank you!

PR #256, opened 2020-04-17
Private Encrypted Channel Docs

daniellevass:
Added docs to help people use private encrypted channels
...

Without docs, people won't know how to use them!
...

### DIFF
--- a/README.md
+++ b/README.md
@@ -623,7 +623,7 @@ with your own encryption master key. There is a
 Update your podfile to include `PusherSwiftWithEncryption` instead of `PusherSwift`.
 
 #### Carthage
-You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift.
+You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift. You will also need to import the `Sodium` framework into your project.
 
 #### Swift Package Manager
 PusherSwiftWithEncryption is not yet compatible with the Swift Package Manager.
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ What else would you want? Head over to one of our example apps:
 
 [CocoaPods](http://cocoapods.org) is a dependency manager for Cocoa projects and is our recommended method of installing PusherSwift and its dependencies.
 
+> If you are looking to use encrypted channels, please read the [specific installation instructions](#cocoapods-1).
+
 If you don't already have the Cocoapods gem installed, run the following command:
 
 ```bash
@@ -96,11 +98,11 @@ $ pod install
 
 Also you'll need to make sure that you've not got the version of PusherSwift locked to an old version in your `Podfile.lock` file.
 
-If you are looking to use encrypted channels, please read the [specific installation instructions](#cocoapods-1).
-
 ### Carthage
 
 [Carthage](https://github.com/Carthage/Carthage) is a decentralized dependency manager that automates the process of adding frameworks to your Cocoa application.
+
+> If you are looking to use encrypted channels, please read the [specific installation instructions](#carthage-1).
 
 You can install Carthage with [Homebrew](http://brew.sh/) using the following command:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -592,6 +594,41 @@ PusherChannel *myPrivateChannel = [pusher subscribeWithChannelName:@"private-my-
...
 
+### Private encrypted channels [BETA]
+
+Similar to Private channels, you can also subscribe to a
+[private encrypted channel](https://pusher.com/docs/channels/using_channels/encrypted-channels).
+This library now fully supports end-to-end encryption. This means that only you and your
+connected clients will be able to read your messages. Pusher cannot decrypt them.
+
+To use this feature, you will need to update your dependencies to use
+`PusherSwiftWithEncryption` instead of `PusherSwift` using whichever package manager you were using originally.
```
TomKemp, line 605: I don't think it is currently compatible with Swift Package Manager. Also, if they are using CocoaPods, then they'll have to import a different Pod which it might be worth explaining in the installation section

README.md
```diff
@@ -592,6 +594,41 @@ PusherChannel *myPrivateChannel = [pusher subscribeWithChannelName:@"private-my-
...
+
+```objc
+PusherChannel *privateEncryptedChannel = [pusher subscribeWithChannelName:@"private-encrypted-my-channel"];
+```
+
+There is also an optional callback in the connection delegate when you can listen for
+any failed decryption events:
+
+```swift
+optional func failedToDecryptEvent(eventName: String, channelName: String, data: String?)
```
TomKemp, line 629: There is a section that explains what all the delegate methods are for that should probably also be updated: https://github.com/pusher/pusher-websocket-swift/blob/75f883779173e85fd1f9fbc5b00c326835ad8f1d/README.md#connection-delegate

mdpye:
I have nothing to add to what Tom said :)

README.md
daniellevass: 👍 have added to both the swift and objective c sections :)

README.md
daniellevass: 👍 have updated the instructions for the different package managers - let me know if there's anything else i can better clarify :)

README.md
```diff
@@ -43,6 +44,7 @@ What else would you want? Head over to one of our example apps:
 - [Subscribing to channels](#subscribing)
   - [Public channels](#public-channels)
   - [Private channels](#private-channels)
+  - [Private encrypted channels [BETA]](#private-encrypted-channels)
```
TomKemp, line 47: This link doesn't work

README.md
```diff
@@ -592,6 +606,52 @@ PusherChannel *myPrivateChannel = [pusher subscribeWithChannelName:@"private-my-
...
+### Installation
+
+#### CocoaPods
+Update your podfile to include `PusherSwiftWithEncryption` instead of `PusherSwift`.
+
+#### Carthage
+You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift.
+
+#### Swift Package Manager
+PusherSwiftWithEncryption is not yet compatible with the swift package manager.
```
TomKemp, line 629:
```suggestion
PusherSwiftWithEncryption is not yet compatible with the Swift Package Manager.
```

README.md
```diff
@@ -592,6 +606,52 @@ PusherChannel *myPrivateChannel = [pusher subscribeWithChannelName:@"private-my-
...
+
+#### Limitations
+
+* Is not currently possible to use with tvOS .
```
TomKemp, line 633:
```suggestion
* Is not currently compatible with tvOS.
```

README.md
```diff
@@ -592,6 +606,52 @@ PusherChannel *myPrivateChannel = [pusher subscribeWithChannelName:@"private-my-
...
+with your own encryption master key. There is a
+[demonstration endpoint to look at using nodejs](https://github.com/pusher/pusher-channels-auth-example#using-e2e-encryption).
+
+### Installation
+
+#### CocoaPods
+Update your podfile to include `PusherSwiftWithEncryption` instead of `PusherSwift`.
+
+#### Carthage
+You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift.
```
TomKemp, line 626:
```suggestion
You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift, along with the addition dependency, Sodium.
```

TomKemp: It's also probably work making it clear in the main Carthage installation section at the top that they only need to include Sodium if they are using end-to-end encryption, not if they're just using `PusherSwift`.

daniellevass: Ah. I didn't know this! Have updated to read : `You do not need to change your Cartfile. However, you will need to import the `PusherSwiftWithEncryption` framework into your project, instead of PusherSwift. You will also need to import the `Sodium` framework into your project.`

README.md
daniellevass: Oh. Good spot! thank you - should be fixed now.

daniellevass:
Thanks for the feedback Tom. 

I've added a section in each of the installation instructions to let people know there's more if they're using encrypted channels and linked (correctly) to the docs. I'm not sure if there's a better way in markdown to make it more noticeable - like an info box or something.

PR #259, opened 2020-04-17
Add warning message for wrong PusherSwift version

TomKemp:
Add a warning message that displays if someone tries to subscribe to an encrypted channel in the PusherSwift rather than PusherSwiftWithEncryption


TomKemp:
It turns out those changes didn't work. My testing incorrectly indicated that it worked. The `WITH_ENCRYPTION` flag is only set for the tests, not for the main schemes. I've changed it to use the same target membership technique as the decryption code. Apologies.

PR #260, opened 2020-04-20
Store the decryption key in the channel

TomKemp:
The decryption key was previously stored in the key provider. Keeping the key in a separate structure meant having to add logic to clean up the key after unsubscribing. This could either happen straight away meaning any events in the queue can't be processed, or when the pending messages are complete, which adds complexity. Storing the key in the channel means the key stays around only as long as there is a reference to the channel, and there will be no more references once the queue has been processed (as long as the consuming app has also released its references to the channel).


Sources/PusherEventQueue.swift
```diff
@@ -15,54 +15,65 @@ class PusherConcreteEventQueue: PusherEventQueue {
     // MARK: - Properties
     
     private let eventFactory: PusherEventFactory
-    private let keyProvider: PusherKeyProvider
+    private let channels: PusherChannels
```
mdpye, line 18: You could have reduced the potential for accidental coupling here by extending PusherChannels to implement the PusherKeyProvider protocol, hiding the details (including the Channel objects) from the event queue as before.

Not required, just food for thought.

Sources/PusherEventQueue.swift
```diff
@@ -15,54 +15,65 @@ class PusherConcreteEventQueue: PusherEventQueue {
...
-        if let channelName = channelName {
-            decryptionKey = self.keyProvider.decryptionKey(forChannelName: channelName)
+    private func processEventWithRetries(json: PusherEventPayload, channel: PusherChannel?) {
+        do {
+            try self.processEvent(json: json, channel: channel)
+        } catch PusherEventError.invalidDecryptionKey {
+            // Reload decryption key if we could not decrypt the payload due to the key
+            // Only events on encrypted channels throw this error, which have a channel
+            if let channel = channel {
+                self.delegate?.eventQueue(self, reloadDecryptionKeySyncForChannel: channel)
```
mdpye, line 57: Is it deliberate that we are calling these delegate methods from the event queue thread, or are they actually expecting to be back on the main thread?

TomKemp:
> Looks good! Just to clarify, this doesn't appear to clear the key after unsubscribing? Will that come in a subsequent PR?

No, the key will never be explicitly cleared. The memory will be freed once there are no more references to the channel.

Sources/PusherEventQueue.swift
TomKemp: That specific example is deliberate because it is synchronous to ensure subsequent events aren't processed until we get the response back from the auth endpoint. Don't want to block the main thread waiting for that.

Sources/PusherEventQueue.swift
TomKemp: That's a very good point!

Sources/PusherEventQueue.swift
mdpye: And the implementation of the delegate method is prepared to be invoked on this thread?

TomKemp: It was, but while double-checking it I realised that part of it should be moved to run on the main thread, so I've just pushed an update.

Sources/PusherConnection.swift
```diff
@@ -938,23 +935,23 @@ extension PusherConnection: PusherEventQueueDelegate {
...
+                    error == nil
+                {
+                    channel.decryptionKey = decryptionKey
+                }else{
+                    channel.decryptionKey = nil
+                }
+                group.leave()
+            }
+        }
+        group.wait()
```
mdpye, line 954: Does this group enter and wait not just achieve the same as using `DispatchQueue.main.sync` rather than `.async`?

I don't think you should require a group unless you are waiting on more than one dispatch which you want to run concurrently.

mdpye: Sorry, reading more closely, I see that the group is left in the *callback* which is scheduled in the main thread.

A comment might be in order :)

TomKemp: Comments added 🙂

PR #261, opened 2020-04-21
Upgrade to Swift 5 (for end-to-end encryption release)

TomKemp:
This PR updates the Swift version from 4.2 to 5.0 and also bumps the dependencies.

This is mainly based on @JonathanDowning's PR #237 but I've done this on a separate branch myself so I could go through the migration process to get the change suggestions from the Xcode migration tool for the end-to-end encryption changes.

The PR also fixes a number of build warnings. There are still build warnings for Starscream, some of which may be resolved if we update the Starscream major version. This was not possible for this release because there was a major API refactor in Starscream. The work can be done after our release.


Cartfile
```diff
@@ -1,3 +1,3 @@
-github "ashleymills/Reachability.swift" == 4.3.0
-github "daltoniam/Starscream" == 3.0.6
+github "ashleymills/Reachability.swift" ~> 5.0.0
+github "daltoniam/Starscream" ~> 3.1.0
```
daniellevass, line 2: what's the difference between `=` and `~>` here?

Tests/PusherSwiftWithEncryption-Only/PrivateEncryptedChannelTests.swift
```diff
@@ -158,7 +158,7 @@ class PrivateEncryptedChannelTests: XCTestCase {
     
     func testTwoEventsReceivedBothFailDecryption() {
         // connect with an incorrect shared secret
-        let channel = subscribeToChannel(authData: incorrectSharedSecretAuthData)
```
daniellevass, line 161: ah, that's a good fix! We should have realised that was potentially a warning before - sorry!

PR #262, opened 2020-04-22
Change to encrypted option to useTLS

TomKemp:
Now that the library has end-to-end encryption support, the meaning of the `encrypted` configuration option is less clear. This PR updates the library to be consistent with other libraries with a `useTLS` option. 

pusher-js has a`forceTLS` option rather than `useTLS`. However the behaviour of that library is different. By default it tries a number of strategies in sequence, including ws and wss. The flag forces the library to only use the  TLS encrypted options e.g. wss. In this library, whether or not the connection uses TLS is entirely determined by whether `encrypted`/`useTLS` is set (same as the server SDKs), so `useTLS` seems more appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ let package = Package(
 There are a number of configuration parameters which can be set for the Pusher client. For Swift usage they are:
 
 - `authMethod (AuthMethod)` - the method you would like the client to use to authenticate subscription requests to channels requiring authentication (see below for more details)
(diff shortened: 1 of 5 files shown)


TomKemp:
> As it's a breaking change, perhaps we should start the changelog now, under an `unreleased` header?

I've opened a draft PR on master and have added an entry to the changelog for it: https://github.com/pusher/pusher-websocket-swift/pull/263

PR #264, opened 2020-04-22
End-to-end encryption support

TomKemp:
This PR adds support for end-to-end encryption by introducing a new `PusherSwiftWithEncryption` target.

This PR also:
- Adds "Consumption Tests" to ensure the library can incorporated into projects using the minimum and maximum supported Swift version and deployment targets. Both `PusherSwift` and `PusherSwiftWithEncryption` are tested. Currently, these can only run locally.
- Increases the minimum deployment target for macOS to 10.11 due to the minimum deployment target of the Sodium dependency.
- Upgrades the Swift version to Swift 5.0 and bumps Reachability/Starscream dependencies accordingly.
- Renames the `encrypted` config option to `useTLS` in order make it clear that option isn't related to end-to-end encryption.

There are breaking changes in this PR (the minimum deployment target bump and the config option change) so the changes will be released as a new major version.


Sources/PusherConnection.swift
```diff
@@ -584,22 +594,26 @@ import Starscream
...
+            "data": error.data ?? ""
         ]
-        let event = PusherEvent(jsonObject: json)!
-        DispatchQueue.main.async {
-            // TODO: Consider removing in favour of exclusively using delegate
-            self.handleEvent(event: event)
-        }
+        if let event = try? self.eventFactory.makeEvent(fromJSON: json, withDecryptionKey: nil){
+            DispatchQueue.main.async {
+                // TODO: Consider removing in favour of exclusively using delegate
```
mdpye, line 606: Leftover TODO

Tests/PusherSwiftWithEncryption-Only/PusherEventQueue+DecryptionTests.swift
```diff
@@ -0,0 +1,349 @@
+import XCTest
+
+#if WITH_ENCRYPTION
```
daniellevass, line 3: I guess this doesn't need the `#if WITH_ENCRYPTION` as it's in the `WithEncryption-Only` package?

Tests/PusherSwift-Only/PusherEventFactory+DecryptionTests.swift
```diff
@@ -0,0 +1,262 @@
+@testable import PusherSwift
+import XCTest
+
+class PusherEventFactoryDecryptionTests: XCTestCase {
```
daniellevass, line 4: might be useful to have a comment at the top of this file to mention that this isn't actually testing decryption in the PusherSwift-only package, but checking to ensure the payloads don't get decrypted.

Tests/PusherSwift-Only/PusherEventQueue+DecryptionTests.swift
```diff
@@ -0,0 +1,90 @@
+import XCTest
+
+#if WITH_ENCRYPTION
```
daniellevass, line 3: likewise - this is in the PusherSwift-Only package - so it doesn't need the `#if WITH_ENCRYPTION`?

Tests/PusherSwift-Only/PusherEventFactory+DecryptionTests.swift
TomKemp: 👍 Done

Tests/PusherSwiftWithEncryption-Only/PusherEventQueue+DecryptionTests.swift
TomKemp: Good spot, fixed

Tests/PusherSwift-Only/PusherEventQueue+DecryptionTests.swift
TomKemp: Fixed

README.md
```diff
@@ -205,13 +205,15 @@ public protocol Authorizer {
 where `PusherAuth` is defined as:
 
 ```swift
-public class PusherAuth: NSObject {
+ public class PusherAuth: NSObject {
```
mdpye, line 208:
```suggestion
public class PusherAuth: NSObject {
```

PR #275, opened 2020-06-29
Subscription event is not triggered if data is null.

cocoBavan:
When Subscription Succeeded event was received with no data, the event listeners were not triggered.

Subscription succeeded message was an essential event to listen to.

### DIFF
--- a/Sources/PusherConnection.swift
+++ b/Sources/PusherConnection.swift
@@ -481,12 +481,7 @@ import Starscream
         if let channelName = event.channelName, let chan = self.channels.find(name: channelName) {
             chan.subscribed = true
 
-            guard event.data != nil else {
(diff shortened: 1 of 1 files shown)


danielrbrowne:
@cocoBavan Apologies for the delay in following up on this.

The event`data` payload should be present for all `pusher:subscribe` events (see [here](https://pusher.com/docs/channels/library_auth_reference/pusher-websockets-protocol#-pusher-subscribe-client-pusher-channels-)). It should contain at minimum a `channel`, and optionally an `auth` and `channel_data` strings.

Can you reproduce a scenario where the `event.data` was in fact `nil` for the `pusher:subscribe` event?

devdreamsolution (approved):

danielrbrowne:
I'm closing this issue due to a lack of response. Feel free to re-open it again in the future if you experience the same issue again (or can provide steps to reproduce it).

Issue #277, opened 2020-07-22
Update PusherSwift to use Starscream 4.x.x


JonathanDowning:
### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### Any improvements you suggest

I noticed that Starscream 3.1.1 has a couple of possible dangling pointer bugs in it's compression code, which could well be causing crashes in our app (although I've no direct evidence of this)
In researching the Starscream codebase I noticed it's now on 4.0.2 and looks much improved.

Are there plans to use this in Pusher soon?

wangdu1005:
Why we use PureSwift have to use Starscream in iOS Swift? I never read any Docs of installation have to use it (Starscream). Can someone help me to figure out?

danielrbrowne:
@wangdu1005 The PusherSwift library itself depends on Starscream (this is used for creating the WebSocket connections with the Pusher platform). So it's not really "used" directly by any developers that use the PusherSwift library.

 If you follow the steps in the [Installation](https://github.com/pusher/pusher-websocket-swift#installation) section of the README, you'll see how to integrate the PusherSwift library within your own iOS or macOS application, using Cocoapods, Carthage or Swift Package Manager as your chosen dependency manager.

chickdan:
We also have a handful of crashes that are traced back to Starscream.
`EXC_BAD_ACCESS0x89cc (-0x102e64000 + 4343646668)
Attempted to dereference garbage pointer 0xee8516020.`

danielrbrowne:
To update interested parties:

I have recently joined Pusher and I will be the primary maintainer of this SDK. I am currently working on a branch to move from Starscream 3.1.1 -> 4.x.x . Currently, I have run into an issue related to how we're mocking their `WebSocket` object for our unit tests, which is blocking my progress as I'd like to avoid introducing websocket connections over a live network as part of our unit testing. I've filed an issue here which, if resolved should unblock my progress: https://github.com/daltoniam/Starscream/issues/811

JonathanDowning:
Thanks for the update @danielrbrowne ! :)

JonathanDowning:
@danielrbrowne I've opened this PR to help get this going https://github.com/daltoniam/Starscream/pull/812

danielrbrowne:
Update:

[...]

PR #280, opened 2020-08-21
WIP: Upgrade Starscream from 3.x -> 4.x

danielrbrowne:
This PR updates the version of Starscream in use from 3.x to 4.x

This addresses #277 (and will allow other Websocket related issues to be addressed in the near future).

**NOTES:**
- This PR is currently WIP (and therefore will not pass CI) until https://github.com/daltoniam/Starscream/issues/811 is resolved upstream in Starscream.


Sources/PusherSwift.swift
```diff
@@ -26,7 +26,7 @@ let CLIENT_NAME = "pusher-websocket-swift"
     public init(key: String, options: PusherClientOptions = PusherClientOptions()) {
         self.key = key
         let urlString = constructUrl(key: key, options: options)
-        let ws = WebSocket(url: URL(string: urlString)!)
+        let ws = WebSocket(request: URLRequest(url: URL(string: urlString)!))
```
JonathanDowning, line 29: Hey @danielrbrowne, just a curiosity question.

Have you experimented with using Starscream's native support for `URLSessionWebSocketTask`?

It's enabled by changing this line to 
``` swift
let ws = WebSocket(request: URLRequest(url: URL(string: urlString)!), useCustomEngine: false)
```

It used to be enabled by default but now is an opt-in. It's not clear why this change was made.
However in my testing (albeit limited), the native socket implementation seemed to work rather well.
Would you be opposed to testing the native iOS sockets implementation? 
I don't know about you, but if I was given the choice of using native networking code versus third party, I'd use native every time! 

Hope you're staying safe! Jon

danielrbrowne:
It is likely that this PR will be closed in favour of #289 instead. This should simplify our dependency graph going forward, with the caveat that the change to native WebSockets will necessitate a version bump to 9.x since deployment targets will have to be increased several versions to allow use of the native API. @JonathanDowning this may be relevant to your interests.

Sources/PusherSwift.swift
danielrbrowne: @JonathanDowning The reasoning for avoiding this until now was the fact that this SDK had minimum deployment targets which wouldn't allow for the use of Starscream's support of `URLSessionWebSocketTask`. We have now re-evaluated this blocker, see [this comment](https://github.com/pusher/pusher-websocket-swift/pull/280#issuecomment-696609804) for more info.

danielrbrowne:
Closing for reasons above.

PR #284, opened 2020-08-27
Refactor String literal usage

### LINKED ISSUE #283: Refactor string literals into a String constants struct
String literals spread throughout the codebase should be refactored into a String constants struct (or similar). This will avoid repeated string literals, which are prone to typos and also will make future refactoring much quicker if these literals change in the future.

danielrbrowne:
This PR resolves #283 

- Refactors several String literals that are spread through the codebase into a `Constants` container:
    - API domains / hosts.
    - Channel types
    - Events
    - Event types
    - JSON keys

- Adds `PusherLogger` for consolidating logging of Debug (or other logging level) events via the `PusherDelegate` `debugLog(message: String)` method.


Sources/Constants.swift
```diff
@@ -0,0 +1,56 @@
...
+        static let data             = "data"
+        static let event            = "event"
+        static let hash             = "hash"
+        static let message          = "message"
+        static let presence         = "presence"
+        static let socketId         = "socket_id"
+        static let sharedSecret     = "shared_secret"
+        static let userId           = "user_id"
+        static let userInfo         = "user_info"
+    }
```
TomKemp, line 55: I'm conflicted about having the JSON keys defined as constants. I was concerned that the same keys might be reused across different event types with different semantics, and linking those keys together with the same constants might make the code harder to understand: it might imply two keys of the same name in different events have the same semantics when they don't, or might cause problems if keys were renamed in some events but not others. But after further thought, I think it's ok because I don't think any of those keys currently have different meanings in different events and it's very unlikely that a key would need to be renamed. Furthermore, it seems useful to have these defined as constants to avoid typos. 

It would be nice if we could eventually have some way of defining the structure of the events using something like `Decodable`, but I seem to remember I had problems with that previously because some of the keys have values that can have different types (e.g. `data` can sometimes be a JSON object and other times be a string)

TomKemp (changes requested):
Looks good to me overall. Just a small change: "presence" and "private" are channel types rather than event types.

Sources/Constants.swift
danielrbrowne: Thanks for the comment - a good observation for the future.

RE: `Decodable` events in the future, we'll be able to do this by implementing `init(from decoder: Decoder) throws` directly:
```swift
// Definition
enum StringInt {
    case string(String)
    case int(Int)
}

// Custom Decodable conformance
extension StringInt: Decodable {
    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        if let stringValue = try? container.decode(String.self) {
            self = .string(stringValue)
        } else if let intValue = try? container.decode(Int.self) {
            self = .int(intValue)
        } else {
            // Throw an error
        }
    }
}
```

Sources/Constants.swift
```diff
@@ -0,0 +1,56 @@
...
+            static let subscriptionSucceeded    = "pusher_internal:subscription_succeeded"
+        }
+    }
+
+    enum EventTypes {
+        static let client           = "client"
+        static let pusher           = "pusher"
+        static let pusherInternal   = "pusher_internal"
+        static let presence         = "presence"
+        static let `private`        = "private"
```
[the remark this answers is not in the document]
danielrbrowne: Good catch!

TomKemp (approved):
👍

PR #285, opened 2020-08-27
Reorganise 'Sources' and 'Tests' directories

### LINKED ISSUE #282: Reorganise source files under the 'Sources' and 'Tests' directories
The source files under the 'Sources' and 'Tests' directories could benefit from being organised by either entity type or responsibility to help developers navigate the Xcode project more easily.

danielrbrowne:
This PR resolves #282 

- Reorganises the `Sources` and `Tests` directories to make navigating the project in Xcode much quicker and easier.

Screenshot 2020-08-28 at 09 48 03
Screenshot 2020-09-08 at 17 05 47


TomKemp (commented):
The structure of the 'Sources' directories looks good. I'm wondering if the structure of the 'Tests' directories should more closely follow the 'Sources' directories.

In other languages, if I was looking for unit tests for the class I was working on, I would expect to look in the same directory within 'Tests' to find them. Is that normally the case in Swift?

Our tests are a mix of unit tests and integration tests I think so I guess it would be different for the integration tests.

danielrbrowne:
@TomKemp Fair comment - I'll push a change to make the two directory structures as similar as possible.

TomKemp (approved):
👍

Issue #222, opened 2019-07-08
Library should respect connection closure codes


TomKemp:
The [Pusher Channels protocol documentation](https://pusher.com/docs/channels/library_auth_reference/pusher-websockets-protocol#connection-closure) specifies that client libraries should not re-establish connections unchanged for error codes 4000-4099, which the library doesn't currently respect.

danielrbrowne:
Having had a look into this - I think the fix for this (and the related closure code logic as specified for codes 4100-4199 where the client should back of for at least 1s) can only be done as part of #277.

This is due to the fact that the closure code is not exposed in the 3.1.1 version of Starscream that is currently used. I.e. it's not present as a parameter on the delegate method `websocketDidDisconnect(socket: WebSocketClient, error: Error?)` whereas it _is_ present in `didReceive(event: WebSocketEvent, client: WebSocket)` with an event of `.disconnected(let reason, let code)`.

PR #289, opened 2020-09-21
Manage WebSocket connections natively

### LINKED ISSUE #277: Update PusherSwift to use Starscream 4.x.x

danielrbrowne:
This PR resolves #277 (in so far as it makes the upgrade redundant)

- Adds a `WebSocket` class (and associated `WebSocketConnection` and `WebSocketConnectionDelegate` protocols), which uses the native `URLSessionWebSocketTask` API to configure and manage WebSocket connections.
- Refactors WebSocket-related code to use the new `WebSocket` class.
- Removes the StarScream dependency from the project.
- Adds a `PusherChannelsProtocolCloseCode` enum to define the [custom close codes](https://pusher.com/docs/channels/library_auth_reference/pusher-websockets-protocol#connection-closure) for the Pusher Channels Protocol. (**These codes are currently not respected.** This enum will be leveraged in a future PR addressing #222).

**NOTES:**
- This PR increases the SDK deployment targets to allow use of the `URLSessionWebSocketTask` API on iOS/tvOS 13.0+ and macOS 10.15+ (The intention will be to release this change as part of v9.0.0 of the SDK, and for some time maintain v8.x releases on a `legacy` branch, with critical bug fixes in parallel with the v9.x+ releases on `master`).


Sources/Models/PusherChannelsProtocolCloseCode.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Ping was sent to the client, but no reply was received
+    case pongReplyNotReceived                   = 4201
+
+    /// Client has been inactive for a long time (currently 24 hours)
+    /// and client does not support ping.
+    case closedAfterInactivity                  = 4202
+
+    // 4300 - 4399
+    case clientEventRejectedDueToRateLimit      = 4300
```
TomKemp, line 82: This error won't result in the connection being closed as far as I'm aware, so it wouldn't be received as a close code. Errors 4000 - 4299 result in the connection being closed whereas 4300 - 4399 would just be received as `pusher:error` events over the websocket with the connection left intact I believe.

Sources/Extensions/PusherWebsocketDelegate.swift
```diff
@@ -126,20 +132,18 @@ extension PusherConnection: WebSocketDelegate {
...
-
-    public func websocketDidReceivePong(socket: WebSocketClient, data: Data?) {
-        self.delegate?.debugLog?(message: PusherLogger.debug(for: .pongReceived))
-        resetActivityTimeoutTimer()
+    func webSocketDidReceiveMessage(connection: WebSocketConnection, data: Data) {
+        //
     }
 
+    func webSocketDidReceiveError(connection: WebSocketConnection, error: Error) {
+        self.delegate?.debugLog?(message: PusherLogger.debug(for: .disconnectionWithError,
```
TomKemp, line 146: As far as I can see, this can be called for errors that aren't necessarily disconnections? E.g. failing to send a message, failing to send a ping etc? Or will they only fail if the websocket is disconnected?

Sources/Extensions/PusherWebsocketDelegate.swift
```diff
@@ -55,14 +66,9 @@ extension PusherConnection: WebSocketDelegate {
...
-        // Handle error (if any)
+        // Log the disconnection
 
-        if let error = error {
-            self.delegate?.debugLog?(message: PusherLogger.debug(for: .disconnectionWithError,
-                                                                 context: "Error (code: \((error as NSError).code)): \(error.localizedDescription)"))
-        } else {
-            self.delegate?.debugLog?(message: PusherLogger.debug(for: .disconnectionWithoutError))
-        }
+        self.delegate?.debugLog?(message: PusherLogger.debug(for: .disconnectionWithoutError))
```
TomKemp, line 71: If the server was to disconnect the client with closure code 4200 (reconnect immediately), what would be printed? Just "Websocket is disconnected but no error received"?

danielrbrowne: The Pusher Channels Protocol custom closure codes are currently not respected (see PR description). This is also currently the case in production, and is also a limitation of the `URLSessionWebSocketTask.CloseCode` enum. I've got a WIP implementation using a lower-level native WebSocket API which will address this separately. The `PusherChannelsProtocolCloseCode` enum in this PR is for future-proofing and is currently unused.

Sources/Models/PusherChannelsProtocolCloseCode.swift
danielrbrowne: Good catch, i'll push a change to remove this.

Sources/Extensions/PusherWebsocketDelegate.swift
danielrbrowne: Good point, these aren't necessarily just errors after a disconnection. I'll push a change to reflect this.

Sources/Extensions/PusherWebsocketDelegate.swift
TomKemp: Sorry, my comment wasn't very clear. I was asking because it looks like less information is being logged now than was being logged previously: we're no longer logging the error code or message that was passed to the Starscream `websocketDidDisconnect` delegate function. I was wondering if the same information would now be logged via `websocketDidReceiveError` instead. I was trying to use 4200 as an example because that isn't really an error as such so I wouldn't have expected that to be passed to `websocketDidReceiveError`. 

But I just tried it out with the current SDK and 4200 doesn't seem to come through as a closure code anyway. It seems to come through as a `pusher:error` event.

TomKemp (approved):

Issue #219, opened 2019-05-30
Subscribe/unsubscribe on non-main thread can cause crash, because internal dictionary is not thread-safe


dtirer:
### Steps to reproduce

I'm using version 7.0.0

Unfortunately uncertain of the steps to reproduce, but i have a large number of the following crashes

### Actual behavior

<img width="1119" alt="Screen Shot 2019-05-30 at 2 48 05 PM" src="https://user-images.githubusercontent.com/639812/58656346-2074e280-82ea-11e9-876b-72576a4e5708.png">

TomKemp:
I think I've been able to replicate this issue.

I was able to replicate it by rapidly subscribing and unsubscribing from channels on a thread that isn't the main thread. The stack trace for the exception in your screenshot shows a message being processed from the websocket, which runs on the main thread and involves reading the message's channel from a Dictionary in `PusherChannels.find`. If you call a method that modifies that Dictionary (e.g. `subscribe` or `unsubscribe`) concurrently on a different thread then you can get the above exception. 

I've started working on a fix for this. I'll let you know when I merge it.

dtirer:
@TomKemp i just wanted to follow up on this issue.

PR #294, opened 2020-09-29
Respect Pusher Channels Protocol WebSocket closure codes

### LINKED ISSUE #222: Library should respect connection closure codes

danielrbrowne:
This PR resolves #222 

- Replaces the implementation of `WebSocket` using `Network.framework`, which is a lower-level native API that exposes all received closure codes whereas `URLSessionWebSocketTask` does not.
  - This is covered by the new `WebSocketTests.swift`.
- Adds logic to `WebSocketConnectionDelegate.swift` to explicitly handle the appropriate reconnection attempt cases presented by Pusher Channels Protocol custom closure codes.
  - **N.B: The `autoReconnect` option is now ignored if the closure code is within the 4000-4999 range** (i.e. the behaviour is determined instead by the received [Channels Protocol closure code](https://pusher.com/docs/channels/library_auth_reference/pusher-websockets-protocol#connection-closure)).
  - This is covered by new unit tests in `PusherConnectionDelegateTests`.
- Ensures the Swift and Objective-C example apps compile correctly by updating their iOS minimum deployment target settings.
- Improves Apple Quick Help documentation for `WebSocketConnection` and `WebSocketConnectionDelegate`.

### DIFF
--- a/Sources/Extensions/PusherWebsocketDelegate.swift
+++ b/Sources/Extensions/PusherWebsocketDelegate.swift
@@ -162,36 +162,21 @@ extension PusherConnection: WebSocketConnectionDelegate {
     /// - Parameter strategy: The reconnection strategy for the reconnection attempt.
     internal func logReconnectionAttempt(strategy: PusherChannelsProtocolCloseCode.ReconnectionStrategy) {
 
-        if case .reconnectImmediately = strategy {
-            if reconnectAttemptsMax != nil {
-                let message = PusherLogger.debug(for: .attemptReconnectionImmediately,
-                                                 context: """
-                    (attempt \(reconnectAttempts + 1) of \(reconnectAttemptsMax!))
-                    """)
-                self.delegate?.debugLog?(message: message)
-            } else {
-                let message = PusherLogger.debug(for: .attemptReconnectionImmediately,
-                                                 context: """
-                    (attempt \(reconnectAttempts + 1))
-                    """)
-                self.delegate?.debugLog?(message: message)
-            }
-        } else {
-            let timeInterval = reconnectionAttemptTimeInterval(strategy: strategy)
-            if reconnectAttemptsMax != nil {
-                let message = PusherLogger.debug(for: .attemptReconnectionAfterWaiting,
-                                                 context: """
-                    \(timeInterval) seconds (attempt \(reconnectAttempts + 1) of \(reconnectAttemptsMax!))
-                    """)
-                self.delegate?.debugLog?(message: message)
-            } else {
-                let message = PusherLogger.debug(for: .attemptReconnectionAfterWaiting,
-                                                 context: """
-                    \(timeInterval) seconds (attempt \(reconnectAttempts + 1))
-                    """)
-                self.delegate?.debugLog?(message: message)
-            }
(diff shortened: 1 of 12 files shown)


Sources/Extensions/PusherWebsocketDelegate.swift
```diff
@@ -129,6 +144,83 @@ extension PusherConnection: WebSocketConnectionDelegate {
...
+                self.delegate?.debugLog?(message: message)
+            } else {
+                let message = PusherLogger.debug(for: .attemptReconnectionAfterWaiting,
+                                                 context: """
+                    \(timeInterval) seconds (attempt \(reconnectAttempts + 1))
+                    """)
+                self.delegate?.debugLog?(message: message)
+            }
+        }
+    }
```
TomKemp, line 195: I don't think it's ideal that there are so many ifs in this logging method. In a way it's replicating parts of the reconnection logic above.

Not too much of an issue but it might be possible to refactor.

Sources/Models/WebSocket.swift
```diff
@@ -1,95 +1,115 @@
...
-                           didCloseWith closeCode: URLSessionWebSocketTask.CloseCode,
-                           reason: Data?) {
-        // swiftlint:disable:next force_try
-        let nwCloseCode = try! NWProtocolWebSocket.CloseCode(rawValue: UInt16(closeCode.rawValue))
-        delegate?.webSocketDidDisconnect(connection: self,
-                                         closeCode: nwCloseCode,
-                                         reason: reason)
+    init(url: URL,
+         connectAutomatically: Bool = false,
+         connectionQueue: DispatchQueue = .global(qos: .default)) {
```
TomKemp, line 48: There are known issues with the library caused by concurrent access to dictionaries (e.g. the one in which channels are stored). I have previously advised people to call methods of this SDK on the main dispatch queue as a work around, because Starscream called delegate methods on the main dispatch queue. Concurrent accesses were therefore avoided because all accesses of dictionaries ran on the same dispatch queue. 

If the queue that websocket delegate methods are being called on is changing, perhaps there will start to be more concurrency issues reported. Obviously we would ideally fix those underlying concurrency issues rather than having to worry about which threads people are calling methods on.

I just wanted to highlight that.

TomKemp (commented):
Looks good. I've left some comments. 

I see it's a WIP. Are you intending to add tests now that the reconnection logic is more complicated? Might also be good to add some form of tests for the WebSocket class now that we're using a lower level API. I'm not sure how we could test that though.

Sources/Models/WebSocket.swift
danielrbrowne: Thanks for the heads up, (I assume you're referring to this issue: https://github.com/pusher/pusher-websocket-swift/issues/219 ?).

In that case, I'll change this init method to default to the main queue for consistency with Starscream, and we can address the underlying thread-safety issue separately.

TomKemp (approved):

Issue #150, opened 2017-09-11
When does this library support the watchOS?


gaozhan1991:
I tried to use the PusherSwift on watchOS platform.
But when running the command "pod install", I got an error:
.....
[!] The platform of the target `Carepoint WatchKit App` (watchOS 3.2) is not compatible with `PusherSwift (4.2.0)`, which does not support `watchos`.
.....

hamchapman:
Currently PusherSwift doesn't work with watchOS. It's something that I'm looking into though, as it involves Starscream (the websocket dependency that PusherSwift uses) having support for watchOS.

I'll leave this issue open and update when I have more news.

gaozhan1991:
Thanks for your reply.

hamchapman:
There's no specific date I can give you at the current time. I'll update this issue when I have more information. Thanks!

hamchapman:
watchOS support has been added to Starscream so I'll try to find time to look into adding support for it soon!

gaozhan1991:
Thanks for your effort.

gaozhan1991:
Hi.
I need your help.
This is the code snippet using javascript. 

client = new Pusher('123', 
				{
					encrypted: true,
					authEndpoint: 'xxx.yyy.zzz',
					auth: {
							params: {
							  'setting1': value1,
							  'setting2': value2,
							  'setting3': value3
							},
							headers: {
							  'id': id
							}
					  }
				}
			);

I want the code corresponding to this one using Swift.
Anything would be much appreciated.

hamchapman:
You'd want something like this:

```swift
class AuthRequestBuilder: AuthRequestBuilderProtocol {
    func requestFor(socketID: String, channelName: String) -> URLRequest? {
        let authEndpointComponents = URLComponents(string: "xxx.yyy.zzz")
        authEndpointComponents?.queryItems = [
            "setting1": value1,
            "setting2": value2,
            "setting3": value3
        ]
        let request = NSMutableURLRequest(url: authEndpointComponents!.url!)
        request.httpMethod = "POST"
        request.httpBody = "socket_id=\(socketID)&channel_name=\(channel.name)".data(using: String.Encoding.utf8)
        request.addValue(id, forHTTPHeaderField: "id")
        return request
    }
}


let options = PusherClientOptions(
    authMethod: AuthMethod.authRequestBuilder(authRequestBuilder: AuthRequestBuilder())
)
client = Pusher(key: "123", options: options)
```

[...]

PR #313, opened 2020-11-20
WebSocket reconnection improvements

### LINKED ISSUE #310: No reconnection after error on version 9.0.0 only
### Steps to reproduce

1. Open the app using the Pusher SDK
2. Wait until a network error occurs (examples "Error (code: 1): The operation couldn’t be completed. (Starscream.WSError error 1." or "Error (code: 0): The operation couldn’t be completed. (Network.NWError error 0.)")

### Expected behavior

The connection should recover, and app should be able to receive new messages again

### Actual behavior

The connection never recovers, you won't receive any message after that

This was not the case on version 8, this is the case only with version 9 of the SDK, we had to revert to version 8 just before submitting the app, because it was unusable on all our testers with a medium-quality internet connection. Just rolling back to version 8 (and just that) fixed the issue.

### LINKED ISSUE #311: Many pusher errors with messages lost
### Steps to reproduce

1. Connect to Pusher on an app with medium-quality connexion (about 50Mbps in Wifi, not even a really bad one like 3G on a train)

### Expected behavior

In most cases, connexion should be OK, and all messages should be received

### Actual behavior

Many errors are showing for many of our users. So the connection is stopped, then restored, but meanwhile the user didn't receive some messages. We are 5 employees working remotely, on these 5, 3 of them are having problems with pusher connexion very regularly (messages are lost, from 1% to about 15% of the time). It was the case with users from https://www.usertesting.com/ too, where 2 users out of 5 had the problem.
I have been using pusher websocket JS for a few years on a large scale project (millions of users monthly) without any of these problems, including within a webview included in an iOS app. On this new project using pusher-websocket-swift, the error rate looks really high to me. The enclosed screenshot is for members of our team, 5 people only.
Screenshot 2020-11-10 at 17 40 02
To be noted that users with really nice connexion do never get these errors (we had to work remotely to detect it). But for other ones, the app is nearly impossible to use.

Check if this is really a normal behavior due to bad conditions, or if it's due to a bug in the library. But looking at how often it happens, I would be surprised that it's only due to a bad network. I wouldn't be surprised that it happens, but not so often. Let me know if I can give any additional information, thanks.

danielrbrowne:
This PR resolves #310 and resolves #311:

- Adds reconnection improvements by updating `NWWebSocket` to version 0.5.0, which includes connection viability and migration reporting via new `WebSocketConnectionDelegate` methods.
- Removes [Reachability](https://github.com/ashleymills/Reachability.swift) as a dependency (since this is now redundant as `NWWebSocket` handles reconnections based on network changes directly).
  - In the future, this should allow for watchOS support (see: #150)
- Resets the minimum Xcode version for 'Consumption-Tests' to Xcode 11.0


Sources/Extensions/PusherWebsocketDelegate.swift
```diff
@@ -85,11 +85,27 @@ extension PusherConnection: WebSocketConnectionDelegate {
...
+        } else {
+            self.delegate?.debugLog?(message: PusherLogger.debug(for: .networkConnectionUnviable))
         }
+    }
 
-        attemptReconnect(closeCode: closeCode)
+    public func webSocketDidAttemptBetterPathMigration(result: Result<WebSocketConnection, NWError>) {
+        switch result {
+        case .success:
+            updateConnectionState(to: .reconnecting)
```
TomKemp, line 102: What is the purpose of updating the state to `reconnecting`? Does a "migration" not behave like a normal disconnection?

Sources/Extensions/PusherWebsocketDelegate.swift
```diff
@@ -85,11 +85,27 @@ extension PusherConnection: WebSocketConnectionDelegate {
             return
         }
 
-        if let reachability = self.reachability, reachability.connection == .unavailable {
-            self.delegate?.debugLog?(message: PusherLogger.debug(for: .reconnectionFailureLikely))
+        attemptReconnect(closeCode: closeCode)
+    }
+
+    public func webSocketViabilityDidChange(connection: WebSocketConnection, isViable: Bool) {
```
TomKemp, line 91: Is `viability` the same concept as `reachability`? If so, how come we're not reacting to the `viability` change? Will it take the SDK longer to realise it's disconnected than it did when reachability was used?

Sources/Extensions/PusherWebsocketDelegate.swift
danielrbrowne: Essentially you've hit the nail on the head, `NWWebSocket` will handle a network path migration itself (i.e. only reporting that one has occured via this delegate method). However, that migration will only re-connect to the Pusher server, in order to trigger re-subscription to the channel(s) set on the `PusherConnection`, the "shadow" state machine must be updated to reflect that a connection change has occurred.

I came to this conclusion in testing, future work could include simplifying or refactoring this state machine but for now I've left it untouched since its part of the public API

Sources/Extensions/PusherWebsocketDelegate.swift
danielrbrowne: In a sense yes, `viability` changes are akin to `reachability` changes - however, we must wait for the network path migration to complete (which takes around the same time as previously when using the `Reachability` dependency anyway) before we can then manage Pusher-related tasks around the connection change.

In summary, the overall reconnection to channel(s) takes approximately the same time, with the pros of consolidating the reconnection logic away from this repo and removing a third party dependency

danielrbrowne: For context, if a device has an active Wi-Fi connection which drops, and also has a 4G signal which is migrated to, the order of events would be:

- `webSocketViabilityDidChange(connection:isViable:)` (where `isViable = false`)
- `webSocketDidAttemptBetterPathMigration(result:)` (migrating to a 4G connection)
- `webSocketViabilityDidChange(connection:isViable:)` (where `isViable = true`)

Sources/Extensions/PusherWebsocketDelegate.swift
TomKemp: Makes sense. Thanks for the explanation. Might make sense to include a summary of that in a comment.

Sources/Extensions/PusherWebsocketDelegate.swift
TomKemp: 👍

TomKemp (approved):
Looks good

PR #322, opened 2020-12-10
Refactor some 'if let' statments to 'guard let'

### LINKED ISSUE #321: Replace 'if let' with 'guard let' where appropriate
There are a number of instances where `if let` is being used, where `guard let` is more appropriate. Refactoring where appropriate (e.g. exiting a method early, focussing on the "happy path", etc) would improve code readability and will adhere more closely to Swift best practices.

danielrbrowne:
This PR resolves #321

- Refactors some cases of `if let` where `guard let` would be more appropriate for Swift best practices.

**NOTES:**

- These changes improve readability by where methods can return early, and allowing focus on the "happy path" of the method.


TomKemp (approved):
👍

PR #323, opened 2020-12-11
Refactor 'print()' logging to use 'PusherLogger'

### LINKED ISSUE #198: Unable to remove debug output / Faulty parsing error
### Steps to reproduce

Hi, so in my project I use Pusher and compress socket messages with gzip. Due to that, `598: getPusherEventJSON(from string: String)` in `PusherConnection.swift` constantly falls into ` print("Unable to parse string from WebSocket: \(string)")`. 

While the cause of this fallback might worth noting in a separate issue, I would love to be able to disable all debug prints in the framework to clean up my logs. Unfortunately, with `print` I can only override this globally or redirect to a virtual source, while I would prefer to be able to toggle it on/off by passing a launch argument to the framework. 

### Expected behavior

Being able to toggle on/off debug output from Pusher.

### Actual behavior

The framework (most recent version installed via cocoapods) contains 20+ `print`s which I can't disable, so my logs get flooded with `Unable to parse string from WebSocket` messages, even though all socket messages are parsed successfully. 

Being able to provide launch arguments to enable or disable debug output.

danielrbrowne:
This PR resolves #198

The SDK can now ensures all logging statements are passed to the `PusherDelegate` `debugLog(message:)` method. This means that some logging messages that were previously `print()` statements can now be silenced by users of the SDK.

- Refactors `PusherLogger` to call the `PusherDelegate` `debugLog(message:)` method internally
  - Adds a shared instance of the logger
  - Adds a `delegate` property which is set automatically when the `PusherConnection` `delegate` is set. (This is used to call `debugLog(message:)`).
- Removes all instances of `print()` logging statements.

### DIFF
--- a/Sources/Extensions/PusherWebsocketDelegate.swift
+++ b/Sources/Extensions/PusherWebsocketDelegate.swift
@@ -11,19 +11,20 @@ extension PusherConnection: WebSocketConnectionDelegate {
         - parameter string: The message received over the websocket
     */
     public func webSocketDidReceiveMessage(connection: WebSocketConnection, string: String) {
-        self.delegate?.debugLog?(message: PusherLogger.debug(for: .receivedMessage, context: string))
+        PusherLogger.shared.debug(for: .receivedMessage, context: string)
 
         guard let payload = PusherParser.getPusherEventJSON(from: string),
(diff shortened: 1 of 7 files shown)


TomKemp (approved):
Looks good to me

PR #330, opened 2021-01-08
Mark stale channel object as unsubscribed

Yoshi2889:

The connection does not mark the channel as unsubscribed after unsubscribing. The channels are detached from PusherChannels at this point already.

This change makes sure the channel's subscribed flag is consistent with its actual state.

### DIFF
--- a/Sources/Services/PusherConnection.swift
+++ b/Sources/Services/PusherConnection.swift
@@ -178,6 +178,7 @@ import NWWebSocket
         )
 
         self.channels.remove(name: channelName)
+        chan.subscribed = false
     }
 
     /**


danielrbrowne:
@Yoshi2889 Is there a reproducible bug that is currently occurring without this change? Are you using the latest v9.1.1 release? If not which version are you using?

I'd just like to be able to compare the behaviour (and maybe add a unit test if needed) before I consider merging this change in.

Yoshi2889:
Thanks for the quick response. I haven't encountered any bug that needed this change, however debugging my application got confusing really quick with unsubscribe calls, because the channel claimed to be subscribed however it was nowhere to be found in the Pusher instance.

I think the general idea is to throw away unsubscribed channel objects anyway so I doubt this change should have much if any impact if channels are treated properly.

If I read the code correctly though, if channels are marked as not subscribed they will not attempt to send any events to the server but will instead locally queue, which sounds like wanted behaviour in this case.

danielrbrowne:
Ok, that makes sense on first reading - I'll do some further testing as soon as possible, and let you know my findings. Thanks again for opening the PR.

PR #333, opened 2021-01-22
Mark legacy 'bind(…)' methods as deprecated

### LINKED ISSUE #332: Mark the legacy event 'bind(…)' methods as deprecated
The v7.1.0 release [introduced](https://github.com/pusher/pusher-websocket-swift/pull/227) new `bind(…)` methods which return a `PusherEvent` object via a callback.

The original methods were left untouched for backwards compatibility, but they should now be marked as deprecated (with the intention to remove them altogether in the future 10.0.0 release).

danielrbrowne:
This PR resolves #332 :

- Marks legacy 'bind(…)' methods as deprecated (these will be removed in the future v10.0.0 release).
- Resolves some SwiftLint warnings
  - Also reduces the SwiftLint warning threshold (prevents new warnings being committed in the future).

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,4 +13,4 @@ identifier_name:
 
 # This generates a compiler error if more than this many SwiftLint warnings are present
 # (This threshold can become more restrictive as remaining warnings are resolved via refactoring)
-warning_threshold: 21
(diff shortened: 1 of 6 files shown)


TomKemp (approved):
Looks good to me

PR #335, opened 2021-02-04
Fix missing 'NWWebSocket' in Carthage integration

junpluse:

Insert missing `'NWWebSocket'` into the Carthage integration step in README.

Without that, the following error message will be displayed at runtime:

```
dyld: Library not loaded: @rpath/NWWebSocket.framework/NWWebSocket
  Referenced from: /path/to/PusherSwift.framework/PusherSwift
  Reason: image not found
```


danielrbrowne:
Thanks for raising this PR!

PR #336, opened 2021-02-05
SwiftLint improvements

danielrbrowne:
This PR:

- Enables some SwiftLint rules which are auto-correctable, and are therefore corrected in this PR.


TomKemp (approved):
👍

PR #337, opened 2021-02-08
Source files reorganisation & access level optimisations

danielrbrowne:
This PR:

- Moves various model objects into their own source files
  - This is to help maintainability and readability
- Moves extensions and objects for Objective-C interop into their own source files
- Favours `private` over `fileprivate` access levels where appropriate
- Removes redundant instances of `internal` access level specifiers for types not nested in `open` or `public` types
  - This is because `internal` is the default access level


TomKemp (approved):
👍

PR #338, opened 2021-02-09
Additions to README for Carthage integration when using Xcode 12.0 and above

danielrbrowne:
Carthage has recently released version `0.37.0`, which includes support for building dependencies using XCFrameworks. This improves the experience of building Carthage dependencies using Xcode 12.0 and above, as the previous workaround is no longer required.

This PR:

- Adds additional information and links to the README to the Carthage integration instructions when using Xcode 12.0 and above.
- Updates the example Carthage workaround script to the latest version (taken from the Carthage repo).


ben-pusher (approved):
LGTM

PR #348, opened 2021-03-30
Package.swift: adds dynamic library

AndrewSB:

...

...

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,8 @@ let package = Package(
     name: "PusherSwift",
     platforms: [.iOS("13.0"), .macOS("10.15"), .tvOS("13.0")],
     products: [
-        .library(name: "PusherSwift", targets: ["PusherSwift"])
+        .library(name: "PusherSwift", targets: ["PusherSwift"]),
+        .library(name: "PusherSwift-Dynamic", type: .dynamic, targets: ["PusherSwift"])
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


danielrbrowne:
@AndrewSB Can you add some details to the description regarding the purpose of this PR and the issues it resolves for you and/or your team?

AndrewSB:
hey! this is a workaround for the issue described here https://forums.swift.org/t/swift-packages-in-multiple-targets-results-in-this-will-result-in-duplication-of-library-code-errors/34892

danielrbrowne:
@AndrewSB Our CI pipeline is currently failing due to an issue I've reported to Carthage... In the meantime, I'm curious if you've tried the Xcode 12.5 beta - discussion in that link you posted suggests there may be a fix for this issue that doesn't require a workaround in `Package.swift`.

AndrewSB:
I actually just downloaded it to try! I'll report back to see if I can use the non-patched pusher once I get around to it

danielrbrowne:
@AndrewSB Apologies for the delay in follow-up with this. It seems the CI problems have been resolved now, so that's good! However, I was wondering if you had any success at your end when using Xcode 12.5. I.e. were you able to build using the latest, unmodified v9.2.2 release of our SDK without an issue?

AndrewSB:
Xcode 12.5 doesn't require this, looks like it works as is!

PR #350, opened 2021-04-23
Enable the watchOS deployment target

CityTransit:

Enable the watchOS deployment target for cocoapods.

We recently did a demo where we've ported some of our app's functionality to the Apple Watch.

To accomplish that project we pointed our dependency to the change in this PR, and as far as we can tell the library works fine on watchOS (at least as of #313)

Tested on device and simulator.


danielrbrowne:
@CityTransit Apologies for the delay in reviewing this PR. For the sake of completeness, I'm closing this PR in favour of #356 which also adds watchOS support to the other `.podspec` file, `Package.swift` and the Xcode project file itself.

PR #374, opened 2022-01-21
Prevent connection reconnect after intentionally disconnect

### LINKED ISSUE #373: Connection automatically reconnects after intentionally calling pusher.disconnect()
### Steps to reproduce
VERSION: "9.2.2"

1. Initialize Pusher instance: let pusher = Pusher(key: "APP_KEY")
2. Start connect to Pusher: pusher.connect()
3. Stop connect to Pusher: pusher.disconnect()

### Expected behavior

No connection to Pusher after that

### Actual behavior

After about 2 minutes, pusher automatically reconnects the connection

As I investigated, the reason is because of pongResponseTimeoutTimer is triggered in PusherConnection.swift, so the connect() method is called again.

My suggestion: We can check whether to reconnect again or NOT based on `intentionalDisconnect` instance property of PusherConnection class.

image

quangthai206:

This PR resolves #373 

To prevent connection reconnecting again after the user intentionally calls pusher.disconnect()

### DIFF
--- a/Sources/Services/PusherConnection.swift
+++ b/Sources/Services/PusherConnection.swift
@@ -358,6 +358,10 @@ import NWWebSocket
         socketConnected = false
         connectionEstablishedMessageReceived = false
         socketId = nil
+        
+        guard !intentionalDisconnect else {
+            return
+        }
 
         attemptReconnect()
     }


sonologico (approved):

sonologico:
Thank you for the bringing this up and suggesting a fix!

PR #383, opened 2022-07-15
Update SDK docs to 10.1.0 version

fbenevides:


docs/Classes.html
```diff
@@ -171,7 +171,7 @@ <h4>Declaration</h4>
                         </div>
                       </div>
                       <div class="slightly-smaller">
-                        <a href="https://github.com/pusher/pusher-websocket-swift/tree/9.2.2/Sources/Services/PusherConnection.swift#L7-L933">Show on GitHub</a>
+                        <a href="https://github.com/pusher/pusher-websocket-swift/tree/10.1.0/Sources/Services/PusherConnection.swift#L7-L948">Show on GitHub</a>
```
amad, line 174: Are these docs embedded with the library? Do we need to release this again to include these updates?

amad (approved):

docs/Classes.html
fbenevides: This generated a Github Page (like [this](https://pusher.github.io/pusher-websocket-swift/)). It's currently outdated, so this PR proposes to update it to the newest version

PR #389, opened 2022-07-27
Set automatically a new release version using Github Actions

fbenevides:
* Set automatically a new release version using Github Actions

## Changelog
* [CHANGED] Use Github Actions to set a new release version automatically

### DIFF
--- a/.github/workflows/prepare.yml
+++ b/.github/workflows/prepare.yml
@@ -14,7 +14,7 @@ jobs:
           fetch-depth: 0
       - name: Fetch the lastest release version
         id: fetch_latest_release
-        uses: thebritican/fetch-latest-release@v1
+        uses: thebritican/fetch-latest-release@v2
         with:
           github_token: ${{ secrets.PUSHER_CI_GITHUB_PRIVATE_TOKEN }}
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/prepare.yml
```diff
@@ -0,0 +1,53 @@
...
+          excludes: prerelease, draft
+      - uses: actions/checkout@v2
+        with:
+          repository: pusher/actions
+          token: ${{ secrets.PUSHER_CI_GITHUB_PRIVATE_TOKEN }}
+          path: .github/actions
+      - uses: ./.github/actions/prepare-version-bump
+        id: bump
+        with:
+          current_version: ${{ steps.fetch_latest_release.outputs.release }}
```
amad, line 31: Can't we read the version number from the source code? 
This might cause issues If someone merges a version bump without releasing it.

amad (approved):

.github/workflows/prepare.yml
fbenevides: Yes, we can. Although it's also good to do that, I believe that by reading this information from releases we'll get this info more accurately. In the releases process, we're making sure that Github releases will be created.

Imagine a situation where someone bumps the version twice in the PR.

PR #442, opened 2025-01-24
Replace NWWebsocket with URLSessionWebSocketTask

theolampert:
This removes the `NWWebsocket` client and replaces it with a wrapper around the newer, higher-level `URLSessionWebSocketTask`. There's still a bit of cleanup to do in this branch, and I'm unsure if it's something you'll want to accept, but I am opening a draft for feedback.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
           pod trunk push PusherSwift.podspec
           pod trunk push PusherSwiftWithEncryption.podspec
         env:
-          COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TOKEN }}
+          COCOAPODS_TRUNK_TOKEN: ${{ secrets.PUSHER_CI_COCOAPODS_TOKEN }}
     outputs:
       release_version: ${{ needs.build.outputs.tag }}
 
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,11 +3,11 @@
     "pins": [
       {
         "package": "NWWebSocket",
-        "repositoryURL": "https://github.com/pusher/NWWebSocket.git",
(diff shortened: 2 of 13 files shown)


theolampert:
This issue still exists

aonemd:
Hey @theolampert Thanks for opening the PR 🙌! I would love to test and merge it. And again, I apologize for taking such a long time to pick it up.

I did some research, and found [this part in a Pusher blog post](https://pusher.com/blog/nwwebsocket-a-flexible-native-websocket-client-for-swift-developers/#introduction:~:text=The%20complication%20at,following%20URLSessionWebSocketDelegate%20method%3A) where it's mentioned we couldn't use `URLSessionWebSocketTask` because Pusher protocol defines its own custom codes:

> The complication at Pusher is that for Channels, we define several custom closure codes describing reasons why a connection could have been closed by our servers. The single omission from the URLSessionWebSocketTask API is the fact that it does not support custom closure codes.

Although, from my limited understanding, the [client code in NWWebSocket](Sources/NWWebSocket/Model/Client/NWWebSocket.swift), is NOT really sending any custom codes and is using the `NWProtocolWebSocket.CloseCode` codes which seems it's the same enum as  [URLSessionWebSocketTask.CloseCode](https://developer.apple.com/documentation/foundation/urlsessionwebsockettask/closecode-swift.enum). 

- Is my understanding correct?
- Have you noticed any issues with reconnection behavior?
- Are you using any Pusher features that depend on custom close codes?

theolampert:
Hey @aonemd I never saw that blog post but that explains better why there was custom handling, we haven't noticed anything strange with reconnection but the reasons to not use `URLSessionWebSocketTask` make sense. I'll double check what I did here, it was a while go, and see if its indeed a problem. I saw you opened another PR to address the crashes so I'll take a look at that as well.

PR #459, opened 2026-04-27
Make Pusher reconnect handling idempotent and allow forced disconnect

lukehillonline:
## Summary

This PR changes `PusherSwift` reconnect and disconnect behavior so that failed connection attempts are handled once, reconnect scheduling is deduplicated, and user-initiated disconnect works even while the client is connecting or reconnecting.

The main goal is to make `PusherConnection` behave with repeated transport failures and with the updated `NWWebSocket` teardown behavior in https://github.com/pusher/NWWebSocket/pull/66, without allowing duplicate terminal callbacks or overlapping reconnect timers to interfere with recovery.

## What Changed

- `disconnect()` now works while the connection is `.connected`, `.connecting`, `.reconnecting`, or `.disconnecting`.
- `disconnect()` now invalidates and clears any pending reconnect timer before tearing down the socket.
- `connect()` now resets per-attempt terminal state and clears any pending reconnect timer before starting a fresh attempt.
- Added per-attempt terminal handling via `terminalEventHandledForCurrentAttempt` so each connection attempt is only terminated once.
- Pong-timeout recovery now disconnects the socket, resets connection state, and schedules a reconnect only once per failed attempt.
- WebSocket error handling now ignores duplicate terminal callbacks from the same failed attempt.
- Reconnect scheduling is now deduplicated so only one reconnect timer can be pending at a time.
- Successful `connection_established` events still reset reconnect attempt counters and clear any pending reconnect timer.

## Why

During investigation, `PusherConnection` could react multiple times to the same failed connection attempt through overlapping timeout, disconnect, and websocket error paths. That led to duplicate state transitions, repeated reconnect scheduling, and noisy recovery behaviour.

This change makes reconnect handling the same per attempt and ensures that user-driven disconnect takes precedence over automatic recovery.

## Behavioral Impact

- User-triggered disconnect is now allowed during active reconnect/connecting states.
- Each failed connection attempt is handled once instead of multiple times through competing callbacks.
- Only one reconnect timer can be pending at a time.
- Reconnect behavior is cleaner when the transport repeatedly fails and retries.

### DIFF
--- a/Sources/Extensions/PusherConnection+WebsocketDelegate.swift
+++ b/Sources/Extensions/PusherConnection+WebsocketDelegate.swift
@@ -50,12 +50,16 @@ extension PusherConnection: WebSocketConnectionDelegate {
     public func webSocketDidDisconnect(connection: WebSocketConnection,
                                        closeCode: NWProtocolWebSocket.CloseCode,
                                        reason: Data?) {
-        resetConnection()
-
-        guard !intentionalDisconnect else {
+        if intentionalDisconnect {
+            resetConnection()
             Logger.shared.debug(for: .intentionalDisconnection)
             return
         }
+
+        if terminalEventHandledForCurrentAttempt { return }
+
+        terminalEventHandledForCurrentAttempt = true
+        resetConnection()
 
         // Log the disconnection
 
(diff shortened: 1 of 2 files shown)


Keith-wright (approved):
Clean implementation, the idempotent reconnect handling and extended disconnect state guards are solid. Approving.

lukehillonline:
Thank you.

